PR #1, opened 2015-09-02
Initial implementation

gfontenot:
This extracts the initial implementation of the basic stack, exposing
the following public API:
- `Request` is a protocol for objects that represent a request to an
  API endpoint. These requests can `build` `NSURLRequest` instances, as
  well as `parse` objects from `JSON` to a specific type. There are no
  concrete classes conforming to this protocol provided by the framework.
- `HTTPResponse` is a simple struct that is returned by objects
  conforming to `RequestPerformer`. This object takes optional data as
  well as an `NSURLResponse` instance (should be an instance of
  `NSHTTPURLResponse`) and pulls out the status code. This lets us
  easily check for API errors down the line.
- `RequestPerformer` is a protocol that describes an object that's able
  to perform requests. We supply a default internal
  `NetworkRequestPerformer`, but this can be overridden if needed. We
  don't expose `NetworkRequestPerformer` directly.
- `Result.Type.fromDecoded` is a function added to `Result` that allows
  simple transformation from `Decoded<T>` to `Result<T, NSError>`.
- `APIClient` is the main concrete class of the library. This object
  takes in `Request`s, and hands back a `Result` with either the
  decoded value, or an instance of `NSError`. Ideally, this class would
  be hidden behind more descriptive functions for specific endpoint
  behavior.

### DIFF
--- a/Source/Extensions/Result.swift
+++ b/Source/Extensions/Result.swift
@@ -1,0 +1,12 @@
+import Argo
+import Result
+
+public extension Result {
+  static func fromDecoded<T>(decoded: Decoded<T>) -> Result<T, NSError> {
+    switch decoded {
+    case let .Success(obj): return .Success(obj)
+    case .TypeMismatch, .MissingKey:
+      return .Failure(error(decoded.description))
+    }
+  }
+}
--- a/Source/Models/APIClient.swift
+++ b/Source/Models/APIClient.swift
@@ -1,0 +1,33 @@
+import Foundation
+import Argo
+import Result
+
+public struct APIClient {
+  private let requestPerformer: RequestPerformer
+
+  public init(requestPerformer: RequestPerformer = NetworkRequestPerformer()) {
+    self.requestPerformer = requestPerformer
+  }
+}
+
+public extension APIClient {
+  func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseType, NSError> -> Void) {
+    requestPerformer.performRequest(request.build()) { result in
+      let object = result >>- deserialize >>- { request.parse($0) }
+      completionHandler(object)
+    }
+  }
+}
+
+private func deserialize(response: HTTPResponse) -> Result<JSON, NSError> {
+  guard let data = response.data else {
+    return .Failure(Result<JSON, NSError>.error("No response data"))
+  }
+
+  do {
+    let object = try NSJSONSerialization.JSONObjectWithData(data, options: NSJSONReadingOptions(rawValue: 0))
+    return .Success(JSON.parse(object))
+  } catch let error as NSError {
+    return .Failure(error)
+  }
+}
--- a/Source/Models/HTTPResponse.swift
+++ b/Source/Models/HTTPResponse.swift
@@ -1,0 +1,12 @@
+import Foundation
+
+public struct HTTPResponse {
+  public let data: NSData?
+  public let code: Int
+
+  public init(data: NSData?, response: NSURLResponse?) {
+    self.data = data
+    self.code = (response as? NSHTTPURLResponse)?.statusCode ?? 500
+  }
+}
+
--- a/Tests/Fakes/FakeSession.swift
+++ b/Tests/Fakes/FakeSession.swift
@@ -1,0 +1,28 @@
+import Foundation
+
+class FakeSession: NSURLSession {
+  var providedRequest = NSURLRequest()
+  var dataTask = FakeDataTask()
+
+  private let data: NSData?
+  private let response: NSURLResponse?
+  private let error: NSError?
+
+  var performedRequest: Bool {
+    return dataTask.resumedTask
+  }
+
+  init(data: NSData? = .None, response: NSURLResponse? = .None, error: NSError? = .None) {
+    self.data = data
+    self.response = response
+    self.error = error
+    super.init()
+  }
+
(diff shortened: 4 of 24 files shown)


Tests/Fakes/FakeSession.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+
+class FakeSession: NSURLSession {
+  var providedRequest: NSURLRequest?
+  var dataTask: FakeDataTask?
+  var completionHandler: ((NSData?, NSURLResponse, NSError?) -> Void)?
```
jakecraige: Is this line required? Doesn't look like you're using it anywhere

Source/Models/HTTPResponse.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+public struct HTTPResponse {
+  public let data: NSData?
+  public let code: Int
+
+  public init(data: NSData?, response: NSURLResponse?) {
+    self.data = data
+    self.code = (response as? NSHTTPURLResponse)?.statusCode ?? 500
```
jakecraige, line 9: Is swift able to infer the type here so you don't have to `as? NSHTTPURLResponse` it?

Source/Models/APIClient.swift
```diff
@@ -0,0 +1,30 @@
...
+    requestPerformer.performRequest(request.build()) { result in
+      let object = result >>- deserialize >>- { request.parse($0) }
+      completionHandler(object)
+    }
+  }
+}
+
+private func deserialize(response: HTTPResponse) -> Result<JSON, NSError> {
+  do {
+    let data = response.data ?? NSData()
```
jakecraige: What happens when you default to `NSData()` like this? When would this happen and what's the expected behavior when it does?

Tests/Fakes/FakeSession.swift
gfontenot: Good catch. I don't remember why I originally added this, but it must have been a dead end.

Source/Models/HTTPResponse.swift
gfontenot, line 9: Unfortunately, no. `NSURLResponse` and `NSHTTPURLResponse` are two different classes, and only the `HTTP` version has a `statusCode` property. Pretty dumb.

jakecraige, line 9: Ah I missed that they were different.

Source/Models/APIClient.swift
gfontenot: defaulting to `NSData()` will cause the `NSJSONSerialization` process below to fail. That will cause it to throw an error, in which case we will transform it into a `Failure` and return it. I'm not sure if this belongs here, or maybe on `HTTPResponse`. I feel like it should be here, because it seems reasonable that `HTTPResponse` might not have any actual data, and so should represent that with a proper type instead of hiding it.

jakecraige:
This looks fantastic. I wasn't exactly looking forward to implementing all this when working on a new iOS app.

Source/Models/APIClient.swift
jakecraige: That makes sense. I suppose in this context we're expecting it to return a JSON response to be parsed, so if it doesn't, then something is wrong. 

I agree it seems reasonable that an `HTTPResponse` could not have data associated(204 No Content) but in this context it's a failure. 

All that being said, did you consider providing a descriptive error message stating that there was no content and we expected there to be? 

I'm not sure how helpful a failed json parsing error would be to someone. Especially since we know exactly what happened we can give a pretty message to help the user out.

gfontenot: I _could_ do something like:

``` swift
guard let data = response.data { return .Failure(.error("No response data")) }
```

That might be nicer than this, and would let us continue to use non-optionals inside the body of this function.

Source/Extensions/Result.swift
```diff
@@ -0,0 +1,12 @@
...
+
+public extension Result {
+  static func fromDecoded<T>(decoded: Decoded<T>) -> Result<T, NSError> {
+    switch decoded {
+    case let .Success(obj): return .Success(obj)
+    case let .TypeMismatch(msg): return .Failure(Result.error(msg))
+    case let .MissingKey(msg): return .Failure(Result.error(msg))
+    }
+  }
+}
```
sidraval: newline

Tests/Tests/APIClientSpec.swift
```diff
@@ -0,0 +1,53 @@
+import Swish
+import Quick
+import Nimble
+import Result
+
+class APiClientSpec: QuickSpec {
```
sidraval: `APIClientSpec` I Imagine

sidraval:
This is a thing of beauty :+1:

gfontenot:
@jakecraige @sidraval Updated with a guard and some better error messaging. More :eyes: when you get a chance?

croaky:
> Like Spalding I get the crowd bouncing
> You're boring, you're platinum but I'm the one touring
> Scoring, right down the bassline SWISH
> You'll get dissed and switched like "KISS-FM"

http://genius.com/Funkmaster-flex-krs-one-freestyle-lyrics

bin/setup
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+brew install carthage 2> /dev/null
```
croaky, line 3: Could grab brew_install_or_upgrade and its helper functions from https://github.com/thoughtbot/laptop/blob/master/mac

jakecraige:
LGTM

sidraval:
Slam Dunk

PR #2, opened 2015-09-02
Add default implementation of parse for Decodable

gfontenot:
If our `ResponseType` is already `Decodable`, we can simplify things and
offer a default implementation of `parse`. This should reduce the times
where we just need to do this manually, while still letting us override the
default implementation if we need to do a more complex parsing operation
(like pulling out root keys, etc)


sidraval:
:+1: Can it be unit tested?

e: LGTM

gfontenot:
> Can it be unit tested?

Talked about this in Slack, but it turns out that this won't compile if it's written any other way, so the unit test probably isn't worth it.

PR #3, opened 2015-09-02
Explicitly code sign framework

gfontenot:
Without this it might cause intermittent issues when building with
Carthage.


sidraval:
:+1:

PR #4, opened 2015-09-02
Update Result to fix code signing

gfontenot:


bin/update
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+brew install carthage 2> /dev/null
+carthage update --use-submodules --no-build --no-use-binaries
```
sidraval, line 4: Sup with this? (e: Really just want to know why we need these flags)

gfontenot, line 4: We found with Argo that in a framework dependency, it's best to use submodules and a workspace. This makes it easier to distribute the framework via git submodules if people want to do that. The `--no-use-binaries` flag ensures that we don't accidentally grab binaries for a dependency if they are offered (we only want the source so that we can pull the project into a workspace), and the `--no-build` just saves us a bit of time by avoiding the build process since it isn't needed.

You can see more discussion about this on thoughtbot/Argo#123

sidraval:
:+1:

PR #5, opened 2015-09-02
Make Request extension public

gfontenot:
If this isn't public, the code consuming this lib doesn't know about the
default implementation, and so this does nothing. We need to make this
public in order to let people use it.


sidraval:
:+1:

PR #11, opened 2015-09-16
Perform completion handler on main thread

tonyd256:
This add a simple `dispatch_async` to the `APIClient` so that it calls the
completion handler on the main thread. Tests have been updated to handle async.


gfontenot:
dope. :sheep:

PR #12, opened 2015-09-18
Update Argo branch

sidraval:
Argo's Swift 2 branch is now master

### DIFF
--- a/Source/Extensions/Result.swift
+++ b/Source/Extensions/Result.swift
@@ -4,9 +4,10 @@ import Result
 public extension Result {
   static func fromDecoded<T>(decoded: Decoded<T>) -> Result<T, NSError> {
     switch decoded {
-    case let .Success(obj): return .Success(obj)
-    case .TypeMismatch, .MissingKey:
-      return .Failure(error(decoded.description))
+    case let .Success(obj):
+      return .Success(obj)
+    case let .Failure(decodedError):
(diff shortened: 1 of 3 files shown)


gfontenot:
:100: merge away

Source/Extensions/Result.swift
```diff
@@ -4,9 +4,10 @@ import Result
 public extension Result {
   static func fromDecoded<T>(decoded: Decoded<T>) -> Result<T, NSError> {
     switch decoded {
-    case let .Success(obj): return .Success(obj)
-    case .TypeMismatch, .MissingKey:
-      return .Failure(error(decoded.description))
+    case let .Success(obj):
+      return .Success(obj)
+    case let .Failure(decodedError):
+      return .Failure(error(decodedError.description))
```
gfontenot, line 10: since `decodeError` is an `ErrorType`, it can just be cast to `NSError`:

``` swift
return .Failure(decodedError as NSError)
```

_but_ we can actually just improve the overall type here and make this return a `Result<T, DecodeError>`. I only used `NSError` because there _wasn't_ a concrete `ErrorType` to use initially.

sidraval, line 10: Doing this will cause us to have to change `NSError` to `DecodeError` all the way down (e.g. in the `Request` protocol itself.) We're going to merge with the cast to `NSError` and revisit at a later date.

PR #14, opened 2015-09-25
Add default parse method for Decodable collections

gfontenot:
We added a default implementation for parse if the response type was
`Decodable`, but weren't able to do the same for collections of decodable
types, because of how Argo implemented the Array extension. With
thoughtbot/Argo#243, we can actually extend Request to automatically parse
ResponseTypes that are CollectionTypes whose elements are themselves
Decodable.

We should point Argo back at master once that PR gets merged.


sidraval:
Lookin' gooooood

PR #15, opened 2015-09-25
Check status code before parsing response

sidraval:
Check that the status code is between 200 and 299 before parsing JSON. Return an appropriate error otherwise.

### DIFF
--- a/Source/Models/APIClient.swift
+++ b/Source/Models/APIClient.swift
@@ -20,8 +20,23 @@ public extension APIClient {
 }
 
 private func deserialize(response: HTTPResponse) -> Result<JSON, NSError> {
+  switch response.code {
+  case 200...299:
+    return parseJSON(response)
+  case 300...399:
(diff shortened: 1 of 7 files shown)


Source/Models/APIClient.swift
```diff
@@ -20,8 +20,23 @@ public extension APIClient {
 }
 
 private func deserialize(response: HTTPResponse) -> Result<JSON, NSError> {
+  switch response.code {
+  case 200...299:
+    return parseJSON(response)
+  case 300...399:
+    return .Failure(Result<JSON, NSError>.error("Multiple choices: \(response.code)"))
```
gfontenot, line 27: Can you not omit the `Result<JSON, NSError>` bit here?

gfontenot:
Makes sense to me. :+1:

gfontenot:
@thoughtbot/ios any thoughts here?

tonyd256:
looks good

dazmuda:
Good stuff. 202!

jnutting:
Good!

Issue #17, opened 2015-10-02
Have APIClient conform to a protocol


gfontenot:
This would be useful for testing purposes.

PR #18, opened 2015-10-02
Move to Client protocol, which APIClient implements

sidraval:
In response to https://github.com/thoughtbot/Swish/issues/17

@gfontenot Not sure if this is what you had in mind, but it makes sense to me!

### DIFF
--- a/Source/Models/APIClient.swift
+++ b/Source/Models/APIClient.swift
@@ -10,8 +10,8 @@ public struct APIClient {
   }
 }
 
-public extension APIClient {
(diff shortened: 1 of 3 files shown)


gfontenot:
I think this is more than we need to do. I don't think it's worth providing default implementations, or exposing the dependency on a request performer. I think `Client` should just be a simple protocol that has the `performRequest` method on it. The rest is fairly specific to an API client, and wouldn't need to be shared.

sidraval:
@gfontenot Ah, yeah I did that a bit blindly.

gfontenot:
No worries, I gave 0 insight into what I was exactly thinking.

This is dope :+1:

PR #20, opened 2015-10-04
Make NetworkRequestPerformer public.

### LINKED ISSUE #19: Make NetworkRequestPerformer public
In an app I'm working on I'd love to subclass this and still be able to call the default implementation. Since it's not exposed I'm unable to do that. 

Happy to open a PR if this is something y'all think would be beneficial.

paulyoung:
This pull request makes `NetworkRequestPerformer` public, as discussed in #19.

I think I'm right in also making the extension public.

fixes #19

### DIFF
--- a/Source/Models/NetworkRequestPerformer.swift
+++ b/Source/Models/NetworkRequestPerformer.swift
@@ -4,7 +4,7 @@ import Result
 public struct NetworkRequestPerformer: RequestPerformer {
   private let session: NSURLSession
 
-  init(session: NSURLSession = NSURLSession.sharedSession()) {
+  public init(session: NSURLSession = NSURLSession.sharedSession()) {
     self.session = session
   }
 }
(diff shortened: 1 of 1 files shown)


Source/Models/NetworkRequestPerformer.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 import Result
 
-struct NetworkRequestPerformer: RequestPerformer {
+public struct NetworkRequestPerformer: RequestPerformer {
   private let session: NSURLSession
 
   init(session: NSURLSession = NSURLSession.sharedSession()) {
```
gfontenot: I think the initializer should probably be public as well.

paulyoung: Yep, fixed. I can squash if you prefer.

gfontenot:
Thanks @paulyoung!

PR #24, opened 2015-10-19
Add Xcode SCM blueprint files to .gitignore

paulyoung:
These were showing up as untracked files in my working directory.

Normally I'd use https://github.com/github/gitignore/blob/master/Swift.gitignore but there were some differences here. Happy to make further changes to that effect.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 *.pbxuser
 xcuserdata
 *.xccheckout
+*.xcscmblueprint
 
 # Build products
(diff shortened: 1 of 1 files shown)


gfontenot:
This is great, thanks. Merged as 435b04c. What are the other differences that we're missing? If they are significant, I'm happy to include them.

paulyoung:
Sort of hard to explain, but probably not a big deal if no one is complaining :trollface:

PR #26, opened 2015-10-20
Add default protocol implementation for parse() where ResponseType is Void

paulyoung:
I found that I needed this today when requests to a particular API endpoint returned a `204` response.

I tried implementing a more meaningful test but couldn't figure out a way that allowed comparison of `()`, `Result<Void, NSError>.Success()`, or its `value`.

### DIFF
--- a/Source/Protocols/Request.swift
+++ b/Source/Protocols/Request.swift
@@ -19,3 +19,9 @@ public extension Request where ResponseType: CollectionType, ResponseType.Genera
     return .fromDecoded(ResponseType.decode(j))
   }
 }
+
+public extension Request where ResponseType == Void {
+  func parse(j: JSON) -> Result<ResponseType, NSError> {
+    return .Success()
+  }
+}
--- a/Source/Protocols/Request.swift
+++ b/Source/Protocols/Request.swift
@@ -2,6 +2,8 @@ import Foundation
 import Argo
 import Result
 
+public typealias EmptyResponse = Void
+
 public protocol Request {
   typealias ResponseType
   func build() -> NSURLRequest
@@ -20,7 +22,7 @@ public extension Request where ResponseType: CollectionType, ResponseType.Genera
   }
 }
 
-public extension Request where ResponseType == Void {
+public extension Request where ResponseType == EmptyResponse {
   func parse(j: JSON) -> Result<ResponseType, NSError> {
     return .Success()
   }
--- a/Tests/Tests/RequestSpec.swift
+++ b/Tests/Tests/RequestSpec.swift
@@ -63,8 +63,7 @@ class RequestSpec: QuickSpec {
         context("when the ResponseType is an EmptyResponse") {
           it("should have a nil Result value") {
             let request = EmptyResponseRequest()
-            let json = JSON.parse([])
-            let result = request.parse(json)
+            let result = request.parse(.Null)
             
             switch result {
             case .Success:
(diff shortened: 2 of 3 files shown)


gfontenot:
Ha, nice. We actually have this same thing in our project, hadn't made a note to extract it.

What do you think about creating an `EmptyResponse` typealias for `Void` to make it a little more explicit? Then you could do `typealias ResponseType = EmptyResponse`. It also means that if we need to improve what an empty response is inside the system, we can just provide it and users won't need to update their code.

paulyoung:
@gfontenot Nice idea. I pushed a change, let me know if you'd prefer it to be somewhere else.

Tests/Tests/RequestSpec.swift
```diff
@@ -51,6 +59,21 @@ class RequestSpec: QuickSpec {
           expect(result.value?.count).to(equal(2))
           expect(result.value?.first?.name).to(equal("giles"))
         }
+        
+        context("when the ResponseType is Void") {
+          it("should have a nil Result value") {
+            let request = VoidResponseRequest()
+            let json = JSON.parse([])
+            let result = request.parse(json)
```
gfontenot: we could probably just do `let result = request.parse(.Null)` here, since the JSON is ignored.

Source/Protocols/Request.swift
```diff
@@ -19,3 +21,9 @@ public extension Request where ResponseType: CollectionType, ResponseType.Genera
     return .fromDecoded(ResponseType.decode(j))
   }
 }
+
+public extension Request where ResponseType == EmptyResponse {
+  func parse(j: JSON) -> Result<ResponseType, NSError> {
+    return .Success()
```
gfontenot, line 27: This is what our implementation looks like too, but is it too naive? I guess we wouldn't get to this point if there were errors along the way?

Tests/Tests/RequestSpec.swift
```diff
@@ -51,6 +59,21 @@ class RequestSpec: QuickSpec {
...
+            
+            switch result {
+            case .Success:
+              XCTAssert(true)
+            default:
+              XCTFail("Unexpected Failure")
+            }
```
gfontenot: I'm tempted to add a simple custom nimble matcher to handle this:

``` swift
import Result

public func beSuccessful<T, E>() -> NonNilMatcherFunc<Result<T, E>> {
  return NonNilMatcherFunc { actual, failure in
    let result = try actual.evaluate()

    switch result {
    case .Success?: return true
    default: return false
    }
  }
}
```

Source/Protocols/Request.swift
paulyoung, line 27: That was my understanding.

Tests/Tests/RequestSpec.swift
paulyoung: Good call. Updated.

Tests/Tests/RequestSpec.swift
paulyoung: Could be useful since it sounds like you're planning on moving that stuff to it's own project anyway.

I can add if you like.

paulyoung: @gfontenot added.

gfontenot:
@paulyoung Want to make sure this is rebased against `master` (to deal with the mac framework), then I can merge it in? This looks great.

paulyoung:
Rebased.

gfontenot:
Pow. Thanks for this, @paulyoung.

paulyoung:
Awesome, thanks.

PR #28, opened 2015-10-21
Enable app extension API only.

paulyoung:
The allows Swish to be used in targets that only support app extension APIs.

Since none of the unavailable APIs are being used, this is a backwards compatible change.

Some more info on this is available here: https://developer.apple.com/library/prerelease/ios/documentation/General/Conceptual/ExtensibilityPG/ExtensionScenarios.html


gfontenot:
Merged as 988bebd

PR #29, opened 2015-10-21
Deployment targets

paulyoung:
This lowers the deployment target for iOS and OS X to 8.0 and 10.10 respectively.


gfontenot:
Merged as dac1f27

Issue #31, opened 2015-10-21
Issue with serializing EmptyResponse


paulyoung:
`The operation couldn’t be completed. (Cocoa error 3840.)`

Apparently this is a common issue with trying to use `NSJSONSerialization` when there is no data.

![failure](https://cloud.githubusercontent.com/assets/84700/10630530/00675c74-778c-11e5-95db-7030c1b68a8e.png)

I was doing something naive in my tests and didn't pick up on this until trying to use it in the wild.

paulyoung:
Fixed in #32

![giphy](https://cloud.githubusercontent.com/assets/84700/10775092/0604c1b2-7cc3-11e5-900d-d97e0647694f.gif)

PR #32, opened 2015-10-21
Don't attempt to serialize empty response data.

paulyoung:
This addresses the issue described in #31.

### DIFF
--- a/Tests/Fakes/FakeRequest.swift
+++ b/Tests/Fakes/FakeRequest.swift
@@ -30,3 +30,19 @@ struct FakeNullDataRequest: Request {
     }
   }
 }
+
+struct FakeEmptyDataRequest: Request {
+  typealias ResponseType = EmptyResponse
+  
+  func build() -> NSURLRequest {
+    return NSURLRequest(URL: NSURL(string: "http://example.com")!)
+  }
+  
+  func parse(j: JSON) -> Result<EmptyResponse, NSError> {
+    switch j {
+    case JSON.Null: return Result.Success(())
+    default:
+      return Result.Failure(Result<EmptyResponse, NSError>.error())
+    }
+  }
+}
--- a/Tests/Tests/APIClientSpec.swift
+++ b/Tests/Tests/APIClientSpec.swift
@@ -66,6 +66,20 @@ class APIClientSpec: QuickSpec {
                 expect(result!.value!).toEventually(beVoid())
               }
             }
+            
+            context("when the data is empty") {
+              it("returns a null JSON object") {
+                let request = FakeEmptyDataRequest()
+                let performer = FakeEmptyResponseRequestPerformer()
+                
+                let client = APIClient(requestPerformer: performer)
+                var result: Result<EmptyResponse, NSError>?
+                
+                client.performRequest(request) { result = $0 }
+                
+                expect(result!.value!).toEventually(beVoid())
+              }
+            }
           }
 
           describe("completionHandler") {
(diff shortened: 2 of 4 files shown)


paulyoung:
I added a test and some fakes but there's probably a better way to do that.

Tests/Fakes/FakeRequest.swift
```diff
@@ -30,3 +30,19 @@ struct FakeNullDataRequest: Request {
...
+  }
+  
+  func parse(j: JSON) -> Result<EmptyResponse, NSError> {
+    switch j {
+    case JSON.Null: return Result.Success(())
+    default:
+      return Result.Failure(Result<EmptyResponse, NSError>.error())
+    }
+  }
+}
```
paulyoung, line 48: Not sure what I was thinking here, this is the same as `FakeNullDataRequest`.

Tests/Tests/APIClientSpec.swift
```diff
@@ -66,6 +66,20 @@ class APIClientSpec: QuickSpec {
...
+              it("returns a null JSON object") {
+                let request = FakeEmptyDataRequest()
+                let performer = FakeEmptyResponseRequestPerformer()
+                
+                let client = APIClient(requestPerformer: performer)
+                var result: Result<EmptyResponse, NSError>?
+                
+                client.performRequest(request) { result = $0 }
+                
+                expect(result!.value!).toEventually(beVoid())
```
paulyoung, line 80: Without the `where` clause I added, this will crash. Not ideal.

paulyoung:
@tonyd256 @gfontenot any suggestion on how to improve this?

paulyoung:
Probably going to take another look at this in the morning.

sidraval:
LGTM :+1:

Tests/Fakes/FakeRequest.swift
paulyoung, line 48: I guess I was trying to distinguish between null data and data with zero bytes, even though the implementations are currently the same.

paulyoung:
Thanks @sidraval. It definitely still feels half baked on the testing side to me.

sidraval:
@paulyoung Perhaps I'm missing something, can you clarify what feels half baked?

paulyoung:
I think mainly:
1. There was already a test for null data, but not data with 0 bytes which is the crux of this issue. I added `FakeEmptyDataRequest` to correspond to `FakeNullDataRequest` but they have the same implementation. I feel like maybe they shouldn't and I may have overlooked something there.
2. Similar story for `FakeEmptyResponseRequestPerformer`. It's close to `FakeRequestPerformer` so maybe they could be combined, or it's a sign that some things can be removed from `FakeEmptyResponseRequestPerformer`.
3. The test case I added will crash if there's a regression. I should fix that.

paulyoung:
Although for number 3, the other tests do the same...

gfontenot:
I'm :+1: on this change in general. I'm also OK with merging these tests as they are, and then trying to refactor these fakes a bit later to clean it up. Fixing this issue was in my mental TODO list (that would _really_ be better not in my head), so I think it's a really good change.

gfontenot:
So, basically, @paulyoung if you feel like 80% good with this as it stands, we can merge it and come back to do a refactoring pass later.

paulyoung:
Sounds good!

paulyoung:
@gfontenot good to go? :bow: :pray:

gfontenot:
@paulyoung yup yup! Merged in as 5911762. Thanks!

PR #33, opened 2015-10-23
Temporary fix for broken Result release.

jnutting:
Xcode 7.1 made Result a little broken. Rob has fixed that, but
not yet put out a release, so I'm pointing at a specific commit
(just so we can have something that works) for now.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,2 +1,2 @@
-github "antitypical/Result" "0.6.0-beta.3"
+github "antitypical/Result" "11d2fe2eee3693a0f1b051637a05244c4b36fbfa"
 github "thoughtbot/Argo" "master"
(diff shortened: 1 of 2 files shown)


gfontenot:
Is there a better branch or something that we can/should point at instead of a specific SHA? `master` maybe?

jnutting:
Not sure. I just did the simplest thing I could, as a temporary fix. Maybe master would be better.

gfontenot:
They also just released a new beta, if we want to point to that.

paulyoung:
I'm currently facing this issue with Carthage when trying to use Swish with ReactiveCocoa `v4.0.0-alpha.3`

```
Could not pick a version for github "antitypical/Result", due to mutually incompatible requirements:
    "0.6.0-beta.5"
    "0.6.0-beta.3"
```

Would love to see a version of Swish that points to Result `0.6.0-beta.5` for that reason.

paulyoung:
Anything I can do to help move this along?

Issue #27, opened 2015-10-20
Extract json payload helpers


gfontenot:
We have some helpers for getting/setting JSON payloads on `NSURLRequest`/`NSMutableURLRequest` that we should include here.

PR #34, opened 2015-10-23
Add JSON payload extensions to NSURLRequest

sidraval:
WRT https://github.com/thoughtbot/Swish/issues/27

We've used these on a couple projects now, and they've proven useful! Most commonly used in the `build()` function of the `Request` protocol, for `POST` requests and the like.

### DIFF
--- a/Source/Extensions/NSURLRequest.swift
+++ b/Source/Extensions/NSURLRequest.swift
@@ -1,0 +1,20 @@
+import Foundation
+
+public extension NSURLRequest {
+  var jsonPayload: NSDictionary {
+    let json = HTTPBody.flatMap { try? NSJSONSerialization.JSONObjectWithData($0, options: NSJSONReadingOptions(rawValue: 0)) }
+    return (json as? [String: AnyObject]) ?? [:]
+  }
+}
+
+public extension NSMutableURLRequest {
+  override var jsonPayload: NSDictionary {
+    get {
+      return super.jsonPayload
+    }
+
+    set {
+      HTTPBody = try? NSJSONSerialization.dataWithJSONObject(newValue, options: NSJSONWritingOptions(rawValue: 0))
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


gfontenot:
Worth throwing some tests around this?

Source/Extensions/NSURLRequest.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+extension NSURLRequest {
+  var jsonPayload: NSDictionary {
+    let json = HTTPBody.flatMap { try? NSJSONSerialization.JSONObjectWithData($0, options: NSJSONReadingOptions(rawValue: 0)) }
+    return (json as? [String: AnyObject]) ?? [:]
```
paulyoung, line 6: Just a heads up on this. I ran into some strange issues because I was also using `[:]` as a fallback.

See https://github.com/the-grid/Disc/commit/02b31d935192307749767c8b095aa11f0270e39a and https://github.com/kylef/Mockingjay/issues/32

gfontenot, line 6: Is this on the other side of the equation? I can see this performing issues if the endpoint returns a root array instead of a dict (which we should probably do something about), but I don't know why this would cause wacky NSURL errors?

paulyoung, line 6: > Is this on the other side of the equation?

I think you're right. My mistake.

sidraval:
Once this gets merged, I'll open an issue around the top level array / dictionary problem.

gfontenot:
:+1:

Source/Extensions/NSURLRequest.swift
```diff
@@ -0,0 +1,20 @@
...
+}
+
+public extension NSMutableURLRequest {
+  override var jsonPayload: NSDictionary {
+    get {
+      return super.jsonPayload
+    }
+
+    set {
+      HTTPBody = try? NSJSONSerialization.dataWithJSONObject(newValue, options: NSJSONWritingOptions(rawValue: 0))
```
paulyoung, line 17: I think the issue I described in https://github.com/thoughtbot/Swish/pull/34#discussion_r42904628 applies here if the dictionary is empty, as shown in https://github.com/the-grid/Disc/commit/02b31d935192307749767c8b095aa11f0270e39a.

Maybe we need a `guard` for `newValue.isEmpty`?

cc @gfontenot @sidraval

PR #37, opened 2015-11-06
Add basic README

sidraval:


README.md
```diff
@@ -1,3 +1,148 @@
...
+    return curry(Comment.init)
+      <^> j <| "id"
+      <*> j <| "commentText"
+      <*> j <| "username"
+  }
+}
+
+let request = CommentRequest(id: 1)
+APIClient().performRequest(request) { response in
+  // let response: Result<Comment, NSError>
```
gfontenot: Maybe add this as part of the closure's type sig, instead of as a comment?

README.md
```diff
@@ -1,3 +1,148 @@
...
+}
+```
+
+We declare a struct that conforms to the `Request` protocol, which forces us to:
+
+* Declare a `ResponseType` that must be `Decodable`, which allows Swish to use
```
gfontenot: It doesn't _have_ to be `Decodable`. But if it is, you get `parse` for free.

README.md
```diff
@@ -1,3 +1,148 @@
...
+  Argo to decode the HTTP response into the appropriate model.
+* Declade a `build` function, which tells Swish what endpoint to hit, and what
```
gfontenot: s/Declade/Declare

gfontenot: Maybe just say "Declare a `build` function, which defines the request to perform? It's really not more complicated than that.

gfontenot:
Solid start :+1:

README.md
sidraval: :+1:

README.md
sidraval: Right -- I'm not mentioning `parse` at all in the basic section, so I figured I might as well say it's gotta be decodable, just to gloss over that issue for now. What do you think?

README.md
sidraval: :+1:

README.md
gfontenot: Yeah, saw that after. I think the word "must" is probably a bit more prescriptive that intended.

PR #38, opened 2015-11-19
[WIP] Post notification on errors

gfontenot:
This still needs tests

It's useful in some cases to get a notification when the server encountered
an error. For example, you could have a top-level object listening for
these network notifications that catches them, then either logs them or
displays them to the user.

This doesn't change the fact that the error is still returned as the
failure part of the result. That means that if something needs to happen
locally in response to the error, we don't prevent that.

### DIFF
--- a/Source/Extensions/GCD.swift
+++ b/Source/Extensions/GCD.swift
@@ -1,0 +1,3 @@
+func onMain(f: dispatch_block_t)  {
+  dispatch_async(dispatch_get_main_queue(), f)
+}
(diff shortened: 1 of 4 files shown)


gfontenot:
I'm actually going to break this up into a couple smaller PRs since the notification bit is a little controversial internally.

paulyoung:
It does seem like different kind of concern to me, and simple enough to post these manually (or perhaps by customizing `performRequest`, for convenience).

PR #39, opened 2015-11-19
Clean up the API client a little

gfontenot:
- Add internal helper for performing actions on the main thread. We aren't
  exporting this as part of the public API.
  - Don't use a closure when we can just pass a function.


paulyoung:
Looks good to me :+1:

PR #40, opened 2015-11-20
Improve errors

gfontenot:
- Add Swish specific error creation
  - Add server JSON to the error object

See commit messages for more detail.


tonyd256:
:+1: I had one comment about the `try` in the Result constructor but otherwise good

paulyoung:
Nice.

PR #42, opened 2015-11-23
Xcode bullshit

gfontenot:
¯\_(ツ)_/¯


paulyoung:
:+1: I think I started seeing the same in projects when I started using 7.1

PR #44, opened 2015-12-02
Add return type in README example

kaishin:


README.md
```diff
@@ -29,11 +29,11 @@ for up to date installation instructions.
 ```swift
 import Swish
 
-func baseRequest(url url: String, method: RequestMethod) {
+func baseRequest(url url: String, method: RequestMethod) -> NSURLRequest {
   let url = NSURL(string: url)!
   let request = NSMutableURLRequest(URL: url)
 
-  request.HTTPMethod = method.rawValue
+  return request.HTTPMethod = method.rawValue
```
paulyoung: This looks a bit strange to me, what with having the assignment on the same line as the return statement.

gfontenot: yeah, I don't think this will work. There should be an added `return request` line instead.

kaishin: My bad... fixing.

gfontenot:
Yep, makes sense. :+1:

gfontenot:
@kaishin I'm squashing/merging this

PR #47, opened 2015-12-04
Add the ability to parse error objects

### LINKED ISSUE #46: Add the ability to model errors as well as successes
It'd be nice to be able to model error states if you know what they look like. It'd be really nice to bake that into the API in a way that lets us do that cleanly. We can probably do this by providing a `ResponseError` typealias alongside `ResponseType` and adding a `parseError` function to `Request`.

We'd want to maintain compatibility with the current API (mainly for simplicity), so we should default the error to `NSError` and add a default implementation for `ResponseError` that behaves like #40

Only concern would be: What happens if `parseError` fails? Should it be able to? How do people recover from an Argo parsing failure? Do we care?

gfontenot:
This adds a `transformError` function to `Request`. This acts as a hook
to be able to change the error from the `NSError` used by the lower
level network clients and JSON parsing into a model more suitable for
your application.

By default, we are still going to return `NSError`s, but it's useful to
have a hook for being able to model the network error response if you
know what it looks like.

This maintains 100% backwards compatibility with the existing API.

Fixes #46

### DIFF
--- a/Source/Models/SwishError.swift
+++ b/Source/Models/SwishError.swift
@@ -1,0 +1,26 @@
+import Foundation
+import Argo
+
+public enum SwishError {
+  case SessionError(NSError)
+  case ServerError(code: Int, json: AnyObject)
+  case JSONSerializationError(NSError)
+  case ArgoError(DecodeError)
+}
+
+extension SwishError: ErrorType { }
+
+public extension SwishError {
+  var rawError: NSError {
+    switch self {
+    case let .SessionError(e):
+      return e
+    case let .JSONSerializationError(e):
+      return e
+    case let .ServerError(code, json):
+      return .error(code, json: json)
+    case let .ArgoError(e):
+      return .error(String(e))
+    }
+  }
+}
--- a/Source/Models/SwishError.swift
+++ b/Source/Models/SwishError.swift
@@ -2,10 +2,10 @@ import Foundation
 import Argo
 
 public enum SwishError {
-  case SessionError(NSError)
+  case ArgoError(Argo.DecodeError)
+  case InvalidJSONResponse(NSError)
   case ServerError(code: Int, json: AnyObject)
-  case JSONSerializationError(NSError)
-  case ArgoError(DecodeError)
+  case URLSessionError(NSError)
 }
 
 extension SwishError: ErrorType { }
@@ -13,9 +13,9 @@ extension SwishError: ErrorType { }
 public extension SwishError {
   var rawError: NSError {
     switch self {
-    case let .SessionError(e):
+    case let .URLSessionError(e):
       return e
-    case let .JSONSerializationError(e):
+    case let .InvalidJSONResponse(e):
       return e
     case let .ServerError(code, json):
       return .error(code, json: json)
--- a/Source/Protocols/Request.swift
+++ b/Source/Protocols/Request.swift
@@ -8,30 +8,30 @@ public protocol Request {
   typealias ResponseObject
   typealias ResponseError = NSError
   func build() -> NSURLRequest
-  func parse(j: JSON) -> Result<ResponseObject, NSError>
-  func transformError(error: NSError) -> ResponseError
+  func parse(j: JSON) -> Result<ResponseObject, SwishError>
+  func transformError(error: SwishError) -> ResponseError
 }
 
 public extension Request where ResponseObject: Decodable, ResponseObject.DecodedType == ResponseObject {
-  func parse(j: JSON) -> Result<ResponseObject, NSError> {
+  func parse(j: JSON) -> Result<ResponseObject, SwishError> {
     return .fromDecoded(ResponseObject.decode(j))
   }
 }
 
 public extension Request where ResponseObject: CollectionType, ResponseObject.Generator.Element: Decodable, ResponseObject.Generator.Element.DecodedType == ResponseObject.Generator.Element {
-  func parse(j: JSON) -> Result<[ResponseObject.Generator.Element], NSError> {
+  func parse(j: JSON) -> Result<[ResponseObject.Generator.Element], SwishError> {
     return .fromDecoded(ResponseObject.decode(j))
   }
 }
 
 public extension Request where ResponseObject == EmptyResponse {
-  func parse(j: JSON) -> Result<ResponseObject, NSError> {
+  func parse(j: JSON) -> Result<ResponseObject, SwishError> {
     return .Success()
   }
 }
 
 public extension Request where ResponseError == NSError {
-  func transformError(error: NSError) -> ResponseError {
-    return error
+  func transformError(error: SwishError) -> ResponseError {
+    return error.rawError
   }
 }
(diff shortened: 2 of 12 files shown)


kaishin:
Interesting approach that didn't occur to me before. One tiny nitpick: success type is called `ResponseType` but error type is called `ResponseError` instead of `ResponseErrorType`. If we could rename things I'd probably go with something like `ResponseSuccessType` and `ResponseFailureType` or something like that.

I'm +1 for merging this as is for now.

gfontenot:
Yeah, had the same thought. I was thinking about renaming `ResponseType` to `ResponseObject`.

kaishin:
I like `ResponseObject` but wouldn't that break backward compat?

gfontenot:
Sure would, that's why I haven't changed it. I think that's an OK change to make before we release 1.0 though.

sidraval:
Nice!

gfontenot:
Can't decide if this is the right approach. Any more feedback from @thoughtbot/ios?

gfontenot:
Also ping @paulyoung in case he has thoughts on this

paulyoung:
I think this seems fine. Without trying to do something custom I can't really offer any more feedback right now :+1:

bernerdschaefer:
Does `NSError` contain enough information to transform into a meaningful domain error object?

I understood #46 as allowing users to work with structured error responses in APIs, like those [returned by Heroku's Platform API](https://devcenter.heroku.com/articles/platform-api-reference#errors).

gfontenot:
@bernerdschaefer Not really. I've actually been working on taking another stab at this. Waiting on some other PRs to come through before I can finish.

gfontenot:
Finally got this updated so that we're not using a raw `NSError` for the transformation. I'd love some more eyes on this if anyone has time.

sidraval:
This looks great to me.

Source/Models/SwishError.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+import Argo
+
+public enum SwishError {
+  case SessionError(NSError)
+  case ServerError(code: Int, json: AnyObject)
+  case JSONSerializationError(NSError)
+  case ArgoError(DecodeError)
```
sharplet: Do you think it's a smell that every case here ends in `Error`? It feels like duplication to me.

Source/Models/SwishError.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+import Argo
+
+public enum SwishError {
+  case SessionError(NSError)
+  case ServerError(code: Int, json: AnyObject)
+  case JSONSerializationError(NSError)
```
sharplet: It looks like this case covers both serialization and deserialization. Is there value in separating those out? My first impression of this was to wonder where the deserialization case was, before I read the code and realized it was included here.

Source/Models/SwishError.swift
```diff
@@ -0,0 +1,26 @@
...
+  case ArgoError(DecodeError)
+}
+
+extension SwishError: ErrorType { }
+
+public extension SwishError {
+  var rawError: NSError {
```
sharplet, line 14: What do you think about changing this to `underlyingError: ErrorType`?

sharplet:
This looks pretty good, I've just left a few notes about the design of `SwishError`.

tonyd256:
I don't have more to add than Adam. Good change :+1:

Source/Models/SwishError.swift
gfontenot: Yeah, for sure. Would you just chop off `Error` at the end of the specific cases?

Source/Models/SwishError.swift
gfontenot: I think this is actually _only_ deserialization, not serialization at all. Should probably rename this to make that clear.

Source/Models/SwishError.swift
gfontenot, line 14: I think that probably makes sense. We can then move from `ErrorType -> NSError` for the default `transformError` method. The only thing that wouldn't work for is the `.ServerError` case. Would you still transform that into an `NSError`?

Source/Models/SwishError.swift
sharplet: That's a good start. I think it would be good to then look at each name and consider whether it's clear from just the name where the error occurred.

For example: `case Session(NSError)`. Is that wrapping an error returned from an `NSURLSession` task? It's not entirely clear.

Source/Models/SwishError.swift
sharplet, line 14: Yeah, well I suppose the idea of an "underlying" error doesn't really make sense in that case. The response _is_ the error, there's nothing underlying. I suppose you could make the return type optional? But it depends on how this method is intended to be used.

Source/Protocols/Request.swift
```diff
@@ -6,24 +6,32 @@ public typealias EmptyResponse = Void
 
 public protocol Request {
   typealias ResponseObject
+  typealias ResponseError = NSError
   func build() -> NSURLRequest
-  func parse(j: JSON) -> Result<ResponseObject, NSError>
+  func parse(j: JSON) -> Result<ResponseObject, SwishError>
+  func transformError(error: SwishError) -> ResponseError
```
sharplet, line 12: Is the intended use case for this to allow clients to map `SwishError` directly into their own error types? What if this function instead accepted a function `<E: ErrorType> SwishError -> E`?

sharplet, line 12: This would essentially turn it into a `mapError` function I think

Source/Models/SwishError.swift
gfontenot, line 14: Primarily, this is used to transform `SwishError` -> `NSError` for the default `transformError` implementation.

Source/Models/SwishError.swift
gfontenot: Do you have naming suggestions?

Source/Protocols/Request.swift
gfontenot, line 12: I'm unclear on where users would be able to actually _use_ that method. It'd move the logic outside the request and into some object up above the `APIClient`.

gfontenot, line 12: And yes, that's what this function is intended for. It mimics `parse` but for the error case.

sharplet:
Rather than continue all the various threads, I thought a diff would be a lot clearer. I've pushed my proposed changes here: https://github.com/thoughtbot/Swish/compare/gf-model-errors...as-alt-swish-error.

What I ended up with was that each error case with an underlying error kept the `Error` suffix. But `JSONSerializationError` was renamed to `InvalidJSONResponse` and `SessionError` to `URLSessionError` for clarity.

I also renamed `rawError` to `NSError` (though I'm not 100% convinced about the capitalization).

gfontenot:
I've updated this PR with @sharplet's naming changes. Any other comments?

sharplet:
LGTM 👍

Issue #50, opened 2015-12-10
Cancellable requests


jakecraige:

gfontenot:
This is probably a dependency of #8

gfontenot:
Implemented in #51

gfontenot:
This is possible now that #51 is merged

PR #51, opened 2015-12-19
Return data tasks when performing requests

gfontenot:
Previously, we were throwing away the data tasks created by
NSURLSession. This made the API simpler, but the end result was that
there was no way to cancel a request once it started.

By returning the data task, we can now leave it up to the consumer on
how to handle canceling requests. If they want the old behavior, they
can simply continue to ignore the return result. But if they want, they
can now hold onto the task that is returned, and cancel it or do
whatever they need to do with it.

### DIFF
--- a/Source/Models/APIClient.swift
+++ b/Source/Models/APIClient.swift
@@ -11,8 +11,8 @@ public struct APIClient {
 }
 
 extension APIClient: Client {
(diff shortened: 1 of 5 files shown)


gfontenot:
Fixes #50

sidraval:
Yup :+1:

paulyoung:
Great change.

Issue #48, opened 2015-12-08
JSON Decoding errors should be from the Swish domain


gfontenot:
Right now, we're using `Result`'s error constructor when transforming from `Decoded` -> `Result`. This isn't a big deal, but it might be clearer to use `Swish`'s constructor for consistency.

paulyoung:
Makes sense. I didn't realize that this didn't happen as part of #40.

I wonder if it would be simple enough to support custom error type so that consumers of API clients backed by Swish see a "MyApiClient" related error instead of something they won't recognize like "Swish".

PR #52, opened 2015-12-19
Create JSON errors under the Swish domain

gfontenot:
Previously we were using the error constructor provided from Result when
converting `Decoded`, to `Result`. This resulted in vending `NSError`
instances with the Result domain. This felt wrong for a number of
reasons.

This change moves the construction of these errors to our internal
function, which means the error is now under our domain.

### DIFF
--- a/Source/Extensions/NSError.swift
+++ b/Source/Extensions/NSError.swift
@@ -1,21 +1,34 @@
 import Foundation
 
-public let NetworkErrorJSONKey = "com.thoughtbot.swish.errorJSON"
+private let swishDomain = "com.thoughtbot.swish"
+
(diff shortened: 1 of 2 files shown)


gfontenot:
Fixes #48

gfontenot:
ping @thoughtbot/ios

Source/Extensions/NSError.swift
```diff
@@ -1,23 +1,36 @@
 import Foundation
 
-public let NetworkErrorJSONKey = "com.thoughtbot.swish.errorJSON"
+private let swishDomain = "com.thoughtbot.swish"
```
tonyd256, line 3: capital `S` on swish?

tonyd256:
Don't mind me coming in here SUPER late ... :+1:

PR #53, opened 2015-12-19
Fix warning in `beVoid` matcher

gfontenot:
Previously we weren't being explicit about the fact that this expression
evaluation was expected to return `Void?`. Since this is a fairly unusual
return type, Xcode was throwing warnings about it, which dirtied the
project's build status and was generally noisy.

To fix this, we can be explicit about expecting `Void?`.

While I'm at it, I'll take the time to clean up the implementation a
little.


sidraval:
LGTM :+1:

PR #54, opened 2015-12-19
Remove duplicate test

gfontenot:
The null data test and the empty data test were identical. In addition, the
custom `parse` function in our empty request was duplicated, since we're
providing it for free with an extension on `Request`.

### DIFF
--- a/Tests/Fakes/FakeRequestPerformer.swift
+++ b/Tests/Fakes/FakeRequestPerformer.swift
@@ -34,22 +34,22 @@ class FakeRequestPerformer: RequestPerformer {
 class FakeEmptyResponseRequestPerformer: RequestPerformer {
   let statusCode: Int
   var passedRequest: NSURLRequest?
-  
+
   init(statusCode: Int = 204) {
     self.statusCode = statusCode
   }
-  
+
   func performRequest(request: NSURLRequest, completionHandler: Result<HTTPResponse, NSError> -> Void) -> NSURLSessionDataTask {
     passedRequest = request
     let data = NSData()
-    
+
     var response: NSHTTPURLResponse? = .None
     if let url = request.URL {
       response = NSHTTPURLResponse(URL: url, statusCode: statusCode, HTTPVersion: .None, headerFields: .None)
     }
-    
+
     let result: Result<HTTPResponse, NSError> = .Success(HTTPResponse(data: data, response: response))
-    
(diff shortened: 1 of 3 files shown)


sidraval:
I think @paulyoung had knowingly duplicated this, but I can't remember why?

paulyoung:
This sort of explains it: https://github.com/thoughtbot/Swish/pull/32#discussion_r42902704

gfontenot:
Are we losing coverage by removing this then? I looked through these and didn't think we were, but maybe I was wrong?

tonyd256:
It appears to me that the spec needs to stay but the 2 request fakes could be consolidated. Since `nil` is different than `NSData()`, the 2 fake request performers need to be used in 2 separate specs

gfontenot:
:+1: that makes sense

gfontenot:
Re-tweaked this to keep the test around but removed the need for duplicated requests and request performers.

@sidraval @tonyd256 @paulyoung want to take another look?

jakecraige:
I like how this API with the `enum` turned out. :+1:

Tests/Fakes/FakeRequestPerformer.swift
```diff
@@ -2,55 +2,45 @@ import Foundation
...
+    .dataWithJSONObject(j, options: NSJSONWritingOptions(rawValue: 0))
+}
 
-  func performRequest(request: NSURLRequest, completionHandler: Result<HTTPResponse, NSError> -> Void) -> NSURLSessionDataTask {
-    passedRequest = request
-    let data = returnedJSON.flatMap {
-      try? NSJSONSerialization.dataWithJSONObject($0, options: NSJSONWritingOptions(rawValue: 0))
-    }
+enum ResponseData {
+  case Data(NSData?)
```
tonyd256, line 11: instead of having an Optional NSData case what about an `Empty` case? then remove the optionality from NSData

tonyd256:
One comment about separating the optional data into another case ... otherwise :+1:

Tests/Fakes/FakeRequestPerformer.swift
gfontenot, line 11: I don't think that matters, since we're just passing this through. The rest of the API deals with optional `NSData`, so it feels nice (to me) to have this be explicit.

PR #55, opened 2015-12-19
Rename ResponseType to ResponseObject

gfontenot:
`ResponseType` made sense to me when I initially wrote this, but as time
went on it made less and less sense. With #47, I think it stops making
sense entirely. This renames `ResponseType` to `ResponseObject` in order to
hopefully clarify the intent a bit more.


sidraval:
:+1:

PR #56, opened 2015-12-19
Add inline documentation for all protocols

gfontenot:
This adds inline documentation for all of the public protocols and their
default extensions.

### DIFF
--- a/Source/Protocols/Request.swift
+++ b/Source/Protocols/Request.swift
@@ -2,28 +2,86 @@ import Foundation
 import Argo
 import Result
 
+/**
+  Used to represent a request that doesn't return any JSON.
+*/
 public typealias EmptyResponse = Void
 
+/// An abstract object representing an interaction with a specific endpoint.
 public protocol Request {
+  /**
+    The object that the endpoint is expected to return.
+  */
   typealias ResponseObject
+
+  /**
+    Build an instance of `NSURLRequest` that hits the specific endpoint.
+
+    - returns: A constructed `NSURLRequest`.
+  */
   func build() -> NSURLRequest
+
+  /**
+    Parse the endpoint's response JSON into a model object.
+
+    - parameter j: The `JSON` from the server.
+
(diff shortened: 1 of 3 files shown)


gfontenot:
ping @thoughtbot/ios @thoughtbot/copy-editing

Source/Protocols/Request.swift
```diff
@@ -2,28 +2,86 @@ import Foundation
 import Argo
 import Result
 
+/**
+  Used to represent a request that doesn't return any JSON.
```
sidraval, line 6: Perhaps this should be more generic, i.e. `Used to represent a request that returns an empty response`? Since you don't _necessarily_ have to use Swish to parse JSON.

sidraval:
This is really great :+1:

Source/Protocols/Request.swift
tonyd256, line 6: I would probably try to stay away from past tense ("used" even though this is a weird case) ... might read better to say what it is like this or something: `A type to represent a request that doesn't return any JSON.`

Source/Protocols/Request.swift
```diff
@@ -2,28 +2,86 @@ import Foundation
...
+               `NSError`.
+  */
   func parse(j: JSON) -> Result<ResponseObject, NSError>
 }
 
 public extension Request where ResponseObject: Decodable, ResponseObject.DecodedType == ResponseObject {
+  /**
+    Parse `JSON` response into a `Decodable` object.
+
+    This is a default implemtnation for `parse` that provides basic support for
```
tonyd256, line 39: implementation

tonyd256:
couple comments but otherwise :+1:

sidraval:
@gfontenot I imagine this needs an overhaul now? Should we leave this open?

gfontenot:
@sidraval I'l close this and re-open as smaller PRs the way I did for Argo.

PR #59, opened 2016-02-01
Allow top level arrays in jsonPayload

sidraval:
- `jsonPayload` now returns type `AnyObject?`, expecting the user to
  type cast appropriately

Damian and I were working on this, and came up this.

While `AnyObject?` isn't ideal, our thoughts are that once the end user sets `.jsonPayload = ...`, they probably won't have a reason to fetch/manipulate it afterward.

Another solution we considered was a SwishRequest, that contained both the `NSMutableURLRequest` and a `JSONPayload` enum that had cases for arrays and dictionaries. This added a lot of overhead in code in exchange for no gain, as far as we could tell.

### DIFF
--- a/Source/Extensions/NSURLRequest.swift
+++ b/Source/Extensions/NSURLRequest.swift
@@ -1,14 +1,19 @@
 import Foundation
 
 public extension NSURLRequest {
-  var jsonPayload: NSDictionary {
+  var jsonPayload: AnyObject {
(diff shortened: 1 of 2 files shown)


gfontenot:
I _think_ we can get rid of the optionality here. We're overriding the setter anyway and always returning a value, so it seems like we could keep the API identical but remove the optional type.

gilesvangruisen:
Agree with Gordon re: removing optionality. Otherwise, looks good :+1:

gilesvangruisen:
Looks good :+1:

PR #62, opened 2016-03-10
Remove hard dependence on JSON

sidraval:
- Adds a `var deserializer: Deserializer` to the `Client` protocol.
  
  Default implementation provided by `JSONDeserializer`.
- Adds a `typealias Decoder: Parser` to the `Request` protocol.
  
  Default implementation provided by `Argo.JSON`

### DIFF
--- a/Source/Extensions/NSError.swift
+++ b/Source/Extensions/NSError.swift
@@ -2,7 +2,7 @@ import Foundation
 
 private let swishDomain = "com.thoughtbot.swish"
 
-public let NetworkErrorJSONKey = swishDomain + ".errorJSON"
+public let NetworkErrorDataKey = swishDomain + ".errorData"
 
 extension NSError {
   static func error(message: String, function: String = #function, file: String = #file, line: Int = #line) -> NSError {
@@ -13,11 +13,11 @@ extension NSError {
     return NSError(domain: swishDomain, code: 0, userInfo: info)
   }
 
-  static func error(statusCode: Int, json: AnyObject, function: String = #function, file: String = #file, line: Int = #line) -> NSError {
+  static func error(statusCode: Int, data: AnyObject?, function: String = #function, file: String = #file, line: Int = #line) -> NSError {
     var info = userInfoFor(function, file, line)
 
     info[NSLocalizedDescriptionKey] = messageForStatusCode(statusCode)
-    info[NetworkErrorJSONKey] = json
+    info[NetworkErrorDataKey] = data
 
     return NSError(domain: swishDomain, code: statusCode, userInfo: info)
   }
--- a/Source/Models/APIClient.swift
+++ b/Source/Models/APIClient.swift
@@ -4,46 +4,35 @@ import Result
 
 public struct APIClient {
   private let requestPerformer: RequestPerformer
+  private let deserializer: Deserializer
 
-  public init(requestPerformer: RequestPerformer = NetworkRequestPerformer()) {
+  public init(requestPerformer: RequestPerformer = NetworkRequestPerformer(), deserializer: Deserializer = JSONDeserializer()) {
     self.requestPerformer = requestPerformer
+    self.deserializer = deserializer
   }
 }
 
 extension APIClient: Client {
   public func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, SwishError> -> Void) -> NSURLSessionDataTask {
     return requestPerformer.performRequest(request.build()) { result in
-      let object = result >>- deserialize >>- request.parse
+      let object = result
+        >>- self.validateResponse
+        >>- self.deserializer.deserialize
+        >>- T.ResponseParser.parse
+        >>- request.parse
+
       onMain { completionHandler(object) }
     }
   }
 }
 
-private func deserialize(response: HTTPResponse) -> Result<JSON, SwishError> {
-  let json = parseJSON(response)
-
-  switch (response.code, json) {
-
-  case let (_, .Failure(e)):
-    return .Failure(e)
-
-  case let (200...299, .Success(j)):
-    return .Success(JSON(j))
-
-  case let (code, .Success(j)):
-    return .Failure(.ServerError(code: code, json: j))
+private extension APIClient {
+  func validateResponse(httpResponse: HTTPResponse) -> Result<NSData?, SwishError> {
+    switch httpResponse.code {
+    case (200...299):
+      return .Success(httpResponse.data)
+    default:
+      return .Failure(.ServerError(code: httpResponse.code, data: httpResponse.data))
+    }
   }
-}
-
-private func parseJSON(response: HTTPResponse) -> Result<AnyObject, SwishError> {
-  guard let data = response.data where data.length > 0 else {
-    return .Success(NSNull())
-  }
-
-  let result = materialize(
-    try NSJSONSerialization
-      .JSONObjectWithData(data, options: NSJSONReadingOptions(rawValue: 0))
-  )
-
-  return result.mapError(SwishError.InvalidJSONResponse)
 }
--- a/Source/Models/JSONDeserializer.swift
+++ b/Source/Models/JSONDeserializer.swift
@@ -1,0 +1,34 @@
+import Argo
+import Result
+
+struct JSONDeserializer: Deserializer {
+  func deserialize(data: NSData?) -> Result<AnyObject, SwishError> {
+    let json = self.parseJSON(data)
+
+    return json.analysis(
+      ifSuccess: Result<AnyObject, SwishError>.init,
+      ifFailure: { .Failure(.InvalidJSONResponse($0)) }
+    )
+  }
+}
+
+extension JSON: Parser {
+  public typealias Representation = JSON
+
+  public static func parse(j: AnyObject) -> Result<JSON, SwishError> {
+    return Result(JSON.init(j))
+  }
+}
+
+private extension JSONDeserializer {
+  private func parseJSON(data: NSData?) -> Result<AnyObject, NSError> {
+    guard let d = data where d.length > 0 else {
+      return .Success(NSNull())
+    }
+
+    return Result(
+      try NSJSONSerialization
+        .JSONObjectWithData(d, options: NSJSONReadingOptions(rawValue: 0))
+    )
+  }
+}
--- a/Source/Protocols/Request.swift
+++ b/Source/Protocols/Request.swift
@@ -6,8 +6,10 @@ public typealias EmptyResponse = Void
 
 public protocol Request {
   associatedtype ResponseObject
+  associatedtype ResponseParser: Parser = JSON
+
   func build() -> NSURLRequest
-  func parse(j: JSON) -> Result<ResponseObject, SwishError>
+  func parse(j: ResponseParser.Representation) -> Result<ResponseObject, SwishError>
 }
 
 public extension Request where ResponseObject: Decodable, ResponseObject.DecodedType == ResponseObject {
--- a/Source/Models/SwishError.swift
+++ b/Source/Models/SwishError.swift
@@ -4,7 +4,7 @@ import Argo
 public enum SwishError {
   case ArgoError(Argo.DecodeError)
   case InvalidJSONResponse(NSError)
-  case ServerError(code: Int, json: AnyObject)
+  case ServerError(code: Int, data: NSData?)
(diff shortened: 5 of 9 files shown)


jsom:
I wrote this with Sid this morning as an example of an alternative `Deserializer/Decoder`. This one is an "null parser", just returning the response's body as a string.

``` swift
import Result

public struct TestRequest: Request {
  public typealias ResponseObject = String
  public typealias Decoder = StringParser

  public func build() -> NSURLRequest {
    return NSURLRequest()
  }

  public func parse(j: String) -> Result<String, NSError> {
    return .Success(j)
  }
}

public struct StringParser: Parser {
  public typealias Representation = String

  public static func parse(j: AnyObject) -> String {
    guard let data = j as? String else { return "" }
    return data
  }
}

public struct NullDeserializer: Deserializer {
  public func deserialize<T : Request>(request: T)(httpResponse: HTTPResponse) -> Result<T.Decoder.Representation, NSError> {

    let string = parseString(httpResponse)
    return string.map { T.Decoder.parse($0) }
  }

  private func parseString(response: HTTPResponse) -> Result<AnyObject, NSError> {
    guard let data = response.data where data.length > 0 else {
      return .Success(NSNull())
    }

    return .Success(data)
  }
}

APIClient(requestPerformer: NetworkRequestPerformer(), deserializer: NullDeserializer())
```

gfontenot:
I _really_ like this in concept. A few things stick out to me as maybe being at the wrong level:
- The null data check will need to be duplicated for each deserializer
- The status code check will need to  be duplicated for each deserializer

These seem like things that the deserializer shouldn't need to care about. I feel like the deserializer should really just care about `NSData -> Result<ResponseObject, ResponseError>` and nothing else. I'm not 100%  sure where the right hook for this is though. I think this implementation is _close_, but seems a bit off.

Cartfile.private
```diff
@@ -1,2 +1,2 @@
-github "Quick/Quick" ~> 0.6.0
-github "Quick/Nimble" "v2.0.0-rc.3"
+github "Quick/Quick" ~> 0.9.2
+github "Quick/Nimble" "v4.0.1"
```
gfontenot: Can we change this to using `~> 4.0` instead of pointing at a specific tag? That was only needed for the rc since Carthage ignores those.

sidraval: 👍

gfontenot: I actually think @sharplet did this already on master

Source/Models/APIClient.swift
```diff
@@ -4,34 +4,34 @@ import Result
...
   }
 }
 
 extension APIClient: Client {
   public func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, T.ResponseError> -> Void) -> NSURLSessionDataTask {
     return requestPerformer.performRequest(request.build()) { result in
-      let object = (result >>- deserialize >>- request.parse).mapError(request.transformError)
+
+      let object = (result
+        >>- self.clearResponse
```
gfontenot: `clearResponse` seems like a weird name to me, am I crazy? I initially thought this was a typo for `cleanResponse`.

Source/Models/APIClient.swift
```diff
@@ -4,34 +4,34 @@ import Result
...
+        >>- self.deserializer.deserialize(request)
+        >>- request.parse).mapError(request.transformError)
```
gfontenot: what do you think about dropping the closing paren and `mapError` onto the next line?

Source/Models/APIClient.swift
gfontenot: maybe `validateResponse`?

Source/Models/JSONDeserializer.swift
```diff
@@ -0,0 +1,30 @@
+import Argo
+import Result
+
+struct JSONDeserializer: Deserializer {
+  func deserialize<T : Request>(request: T)(data: NSData?) -> Result<T.Decoder.Representation, SwishError> {
```
gfontenot: Implicit currying syntax is being removed. We should probably update this to use explicit currying.

Source/Models/JSONDeserializer.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    switch json {
+    case let .Failure(e):
+      return .Failure(.InvalidJSONResponse(e))
+    case let .Success(j):
+      return .Success(T.Decoder.parse(j))
+    }
+  }
+
+  private func parseJSON(data: NSData?) -> Result<AnyObject, NSError> {
```
gfontenot, line 24: Can we move this into a private extension?

Source/Protocols/Client.swift
```diff
@@ -4,4 +4,6 @@ import Result
 
 public protocol Client {
   func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, T.ResponseError> -> ()) -> NSURLSessionDataTask
+  func clearResponse(httpResponse: HTTPResponse) -> Result<NSData?, SwishError>
```
gfontenot: oh interesting. I don't think I realized this would be part of the protocol.

Source/Protocols/Client.swift
```diff
@@ -4,4 +4,6 @@ import Result
...
+  var deserializer: Deserializer { get }
```
gfontenot: Can we move this above the methods?

Source/Models/JSONDeserializer.swift
gfontenot: Actually, maybe this _shouldn't_ be curried at all, and we should just curry it at the call site using an internal implementation of `curry`.

Carthage/Checkouts/Argo
```diff
@@ -1 +1 @@
-Subproject commit be127e583674ff8e92653c9bbd0328ab55daebc7
+Subproject commit 9d839beffebee85cac2831d302fa92b509791acf
```
gfontenot: This shouldn't be changed. Maybe run `bin/setup` again?

gfontenot: to be clear, this is an _older_ version of Argo and so is breaking CI

gfontenot:
My only real question of consequence is: should `clearResponse` be part of the `Client` protocol? Why would users care about that?

Source/Models/APIClient.swift
sidraval: 👍

Source/Models/APIClient.swift
sidraval: 👍

Source/Protocols/Client.swift
sidraval: Yeah, I wasn't totally sure. This PR kind of has a theme of "extreme flexibility with sensible defaults" and that's why I chose to add it to the protocol; the default APIClient just passes along 2xx, but other implementers can choose to do whatever they wish... thoughts?

Source/Protocols/Client.swift
sidraval: 👍

Carthage/Checkouts/Argo
sidraval: 👍

Source/Models/JSONDeserializer.swift
sidraval, line 24: 👍

sidraval:
@gfontenot Replying here because the diff got squashed:

Yeah, I wasn't totally sure. This PR kind of has a theme of "extreme flexibility with sensible defaults" and that's why I chose to add it to the protocol; the default APIClient just passes along 2xx, but other implementers can choose to do whatever they wish... thoughts?

gfontenot:
IMO it seems like it'd make more sense to keep `validateResponse` as an implementation detail for now. I honestly see more people creating their own `RequestPerformer`s than `Client`s. That protocol was created almost entirely for testing purposes.

Source/Extensions/NSError.swift
```diff
@@ -13,11 +13,11 @@ extension NSError {
     return NSError(domain: swishDomain, code: 0, userInfo: info)
   }
 
-  static func error(statusCode: Int, json: AnyObject, function: String = #function, file: String = #file, line: Int = #line) -> NSError {
+  static func error(statusCode: Int, data: AnyObject?, function: String = #function, file: String = #file, line: Int = #line) -> NSError {
     var info = userInfoFor(function, file, line)
 
     info[NSLocalizedDescriptionKey] = messageForStatusCode(statusCode)
-    info[NetworkErrorJSONKey] = json
+    info[NetworkErrorJSONKey] = data
```
sharplet: Given that this is a breaking change, maybe rename `NetworkErrorJSONKey`?

Source/Protocols/Request.swift
```diff
@@ -6,8 +6,11 @@ public typealias EmptyResponse = Void
 
 public protocol Request {
   associatedtype ResponseObject
+  associatedtype ResponseError = NSError
+  associatedtype Decoder: Parser = JSON
```
sharplet: Should the associated type also be `Parser`?

Source/Models/JSONDeserializer.swift
```diff
@@ -0,0 +1,37 @@
+import Argo
+import Result
+
+struct JSONDeserializer: Deserializer {
+  func deserialize<T : Request>(request: T) -> (NSData?) -> Result<T.ResponseParser.Representation, SwishError> {
```
gfontenot: So this `request` only exists to act as a hint for what `T` is, right? What if we made `JSONDeserializer` itself generic over `T: ResponseParser` (or would it need to be `Deserializer` that's generic)? I guess that adds the complication that we'd need to move the deserializer instantiation down to the `performRequest` method, which might not be ideal? I don't know. It just always bothers me to have a generic function like this that needs to take an otherwise unused instance of an object to act as a type hint.

sidraval: Yeah, its just so `T.ResponseParser` is concrete. I don't like it either, but moving the deserializer to the `performRequest` level makes the API feel way more complicated, and kind of confuses the point of this PR, because now instead of having different clients for different response parsing, you can have one client that can decode any number of different response types.

Source/Models/JSONDeserializer.swift
```diff
@@ -0,0 +1,36 @@
+import Argo
+import Result
+
+struct JSONDeserializer: Deserializer {
+  func deserialize(data: NSData?) -> Result<AnyObject, SwishError> {
+    let json = self.parseJSON(data)
+
+    switch json {
```
gfontenot: you know, this could probably use `analysis` instead of a switch:

``` swift
return json.analysis(
  ifSuccess(.Success),
  ifFailure { .Failure(.InvalidJSONResponse($0)) } // could obviously write a helper function for this
)
```

might need More Type Info, but just a thought.

Source/Models/APIClient.swift
```diff
@@ -4,46 +4,35 @@ import Result
 
 public struct APIClient {
   private let requestPerformer: RequestPerformer
+  public let deserializer: Deserializer
```
sharplet: Seems inconsistent that one of these is private while the other is public?

Source/Models/JSONDeserializer.swift
sidraval: 👍

Source/Models/APIClient.swift
sidraval: Good point, `requestPerformer` should probably be apart of `Client`

gfontenot: Does that actually matter for the `Client` protocol? Does `deserializer`? Or are these both implementation details of our default `APIClient`. I'd argue the latter.

gfontenot: As in, I'd prefer we _remove_ `deserializer` from the `Client` protocol and make this `private`. If there is a need to make these `public` down the line, that will be super easy, where as marking these as `private` later will be more difficult.

sharplet: Yeah I think making both properties `private` is a good choice for now.

sharplet:
LGTM! Just the one comment: https://github.com/thoughtbot/Swish/pull/62#discussion_r62382578

Source/Models/APIClient.swift
sidraval: After discussion in slack, @gfontenot is right, I'll make both private and remove them from the `Client` protocol.

gfontenot:
Wooo! 👍 🎳 💯 🍨 🌖

Issue #63, opened 2016-03-17
Allow form style HTTP request bodies


sidraval:
Some APIs want POST bodies as form style, base64 encoded strings, i.e. `key1=value1&key2=value2` instead of JSON. Maybe we should have a way to convert a `[String: AnyObject]` dictionary into form-style bodies?

paulyoung:
I think my comments in #49 apply here too, in case that's of any help in implementing this.

sidraval:
Thanks @paulyoung !

Work is in progress here https://github.com/thoughtbot/Swish/tree/sr-form-style-payloads

sidraval:
#64

PR #64, opened 2016-03-20
Allow application/x-www-form-urlencoded request body payloads

sidraval:
The API is basically:

``` swift
urlRequest.formStylePayload = [
  "firstKey": "firstValue",
  "secondKey": "secondValue"
]
```

It will automatically percent-encode characters that are forbidden in the URL (e.g. if you wanted to send the value `value=&%`), and then it will base64 encode the string.

WRT #63

---

This change is [Reviewable](https://reviewable.io/reviews/thoughtbot/swish/64)


sidraval:
Closes #63

Source/Extensions/NSURLQueryItem.swift
```diff
@@ -0,0 +1,41 @@
+public extension NSURLQueryItem {
+  static func initAndEncode(key: String, value: String) -> NSURLQueryItem? {
```
gfontenot: It seems weird that we need this. Does `NSURLQueryItem(name: value:)` not do this automatically?

gfontenot: It _seems_ like we should leave this up to the user. Is that crazy?

Source/Extensions/NSURLQueryItem.swift
```diff
@@ -0,0 +1,41 @@
...
+    guard let key = el[safe: 0]?.stringByRemovingPercentEncoding,
+          let value = el[safe: 1]?.stringByRemovingPercentEncoding
+          else { return accum }
+    return accum.flatMap { $0 + [NSURLQueryItem(name: key, value: value)] }
+  }
+
+  private static func encode(s: String) -> String? {
+    let characterSet = NSMutableCharacterSet()
+    let disallowed = NSCharacterSet.URLQueryAllowedCharacterSet()
+    characterSet.formUnionWithCharacterSet(disallowed.invertedSet)
```
gfontenot: What do you think about moving `invertedSet` to the above line so that `disallowed` is named properly from the jump?

Source/Extensions/NSURLQueryItem.swift
```diff
@@ -0,0 +1,41 @@
...
+    characterSet.addCharactersInString("&=")
```
gfontenot: these aren't already included in the query allowed character set?

Source/Extensions/NSURLQueryItem.swift
```diff
@@ -0,0 +1,41 @@
...
+
+  private static func stringToQueryItem(accum: [NSURLQueryItem]?, el: [String]) -> [NSURLQueryItem]? {
+    guard let key = el[safe: 0]?.stringByRemovingPercentEncoding,
+          let value = el[safe: 1]?.stringByRemovingPercentEncoding
+          else { return accum }
+    return accum.flatMap { $0 + [NSURLQueryItem(name: key, value: value)] }
+  }
+
+  private static func encode(s: String) -> String? {
+    let characterSet = NSMutableCharacterSet()
```
gfontenot: What about `NSMutableCharacterSet(charactersInString: "&=")` here to remove the mutation on line 31?

Source/Extensions/NSURLQueryItem.swift
gfontenot: it's too bad there isn't a pure version of this that returns a new set

Source/Extensions/NSURLQueryItem.swift
```diff
@@ -0,0 +1,41 @@
...
+    let disallowed = NSCharacterSet.URLQueryAllowedCharacterSet()
+    characterSet.formUnionWithCharacterSet(disallowed.invertedSet)
+    characterSet.addCharactersInString("&=")
+
+    return s.stringByAddingPercentEncodingWithAllowedCharacters(characterSet.invertedSet)
+  }
+}
+
+private extension Array {
```
gfontenot: Can we move this out to its own file and make it internal?

Source/Extensions/NSURLRequest.swift
```diff
@@ -22,4 +34,32 @@ public extension NSMutableURLRequest {
...
+        HTTPBody = queryString.dataUsingEncoding(NSUTF8StringEncoding)
+      }
+    }
+  }
+}
+
+private extension CollectionType where Generator.Element == NSURLQueryItem {
+  var queryString: String {
+    return sort { $0.name < $1.name }
+    .reduce([]) { accum, el in
```
gfontenot: I feel like this should be indented

gfontenot:
I have some questions that expose my general lack of familiarity with `NSURLQueryComponents`, but I _think_ the rest of this makes sense.

Source/Extensions/NSURLQueryItem.swift
```diff
@@ -0,0 +1,41 @@
...
+    guard let data = data else { return [] }
+    let result = String(data: data, encoding: NSUTF8StringEncoding)
+      .flatMap { $0.componentsSeparatedByString("&") }?
+      .map { $0.componentsSeparatedByString("=") }
+      .reduce([], combine: stringToQueryItem)
+
+    return result ?? []
+  }
+
+  private static func stringToQueryItem(accum: [NSURLQueryItem]?, el: [String]) -> [NSURLQueryItem]? {
```
tonyd256: does this have to take optional `accum` and return an optional? I think this function could remove those and be fine.

Source/Extensions/NSURLRequest.swift
```diff
@@ -10,6 +10,18 @@ public extension NSURLRequest {
     let jsonArray = json as? [AnyObject]
     return jsonArray ?? [:]
   }
+
+  var formStylePayload: [String: String]? {
```
tonyd256: `formStylePayload` doesn't seem right ... if we look at the encoding types, you're dealing with `application/x-www-form-urlencoded` right? so maybe `urlEncodedPayload`?

tonyd256: Does Swish set the `Content-Type` header? Can't find it in the repo

gfontenot: No, that's up to each individual request and is a consumer level concern.

Source/Extensions/NSURLQueryItem.swift
```diff
@@ -0,0 +1,41 @@
...
+
+    return NSURLQueryItem(name: encodedKey, value: encodedValue)
+  }
+
+  static func decode(data: NSData?) -> [NSURLQueryItem] {
+    guard let data = data else { return [] }
+    let result = String(data: data, encoding: NSUTF8StringEncoding)
+      .flatMap { $0.componentsSeparatedByString("&") }?
+      .map { $0.componentsSeparatedByString("=") }
+      .reduce([], combine: stringToQueryItem)
```
dazmuda: I think this could just be a map, no?
You're taking an array of strings and turning them into an array of NSURLQueryItems.

dazmuda: Also is there maybe a way to use NSURLComponents to split up this URL rather than treating it as a string? Like NSURLComponents(result).queryItems or some such?

sharplet:
After some investigation, it looks like the current implementation will reject some valid query strings. Specifically, the `=<value>` portion of a parameter is completely optional, meaning the string `novalue` is perfectly valid.

Using `[String: String]`, we're not going to be able to model query strings according to spec. In practice this might be fine, but as a general-purpose networking library this doesn't feel quite right to me.

What about switching to `[URLQueryItem]`? If we did make that change, it would probably make sense to remove the percent-encoding step we're currently performing, and use `URLComponents` to encode the query string. We could potentially expose `URLQueryItem.encode` and `URLQueryItem.decode` for users who want to opt in to the extra level of percent-escaping.

gfontenot:
I have no problem with that

sharplet:
Ready for re-review 👍

Source/Extensions/URLRequest.swift
```diff
@@ -0,0 +1,27 @@
...
+  var jsonPayload: Any {
+    get {
+      return httpBody.flatMap { try? JSONSerialization.jsonObject(with: $0) } ?? [:]
+    }
+    set {
+      httpBody = try? JSONSerialization.data(withJSONObject: newValue)
+    }
+  }
+
+  @available(macOSApplicationExtension 10.10, *)
```
tonyd256, line 13: is this not possible on iOS or does this mean to restrict macOS and not restrict iOS?

Source/Extensions/URLRequest.swift
```diff
@@ -0,0 +1,27 @@
...
+    get {
+      var components = URLComponents()
+      components.query = httpBody.flatMap { String(data: $0, encoding: .utf8) }
+      return components.queryItems ?? []
+    }
+
+    set {
+      var components = URLComponents()
+      components.queryItems = newValue
+      httpBody = components.query!.data(using: .utf8)
```
tonyd256: should this be a `!`? ... can the user potentially send something to this that causes this to crash?

tonyd256: The JSON version is setting this with an optional so maybe it's ok to do that here too?

Source/Extensions/URLRequest.swift
sharplet: This only introduces a minimum version restriction for macOS application extensions. The `*` means that all other platforms can use it.

Source/Extensions/URLRequest.swift
sharplet: Unless I'm mistaken, I believe that setting `queryItems` should always update `query`. My assumption is that if you can construct a `URLQueryItem`, you can always get a valid query string. The return type of this expression is still optional, but then maybe it would be fine to just be more conservative and use optional chaining here?

tonyd256: if you can simply replace the `!` with a `?` and it all still works, then I think that's better.

PR #68, opened 2016-04-22
Don't force completion blocks to be scheduled on the main queue

sharplet:
(This PR depends on #67.)

Different request performers are allowed to have different semantics when it comes to the queue that results are delivered on. If a performer already happens to deliver results on the main queue, `APIClient` is injecting latency by scheduling an unnecessary async call. If a performer delivers results on a background queue (i.e., `NetworkRequestPerformer`), and there's further processing that would benefit from occurring in the background (e.g., image processing or decoding a large object), callers must dispatch back onto a background queue to avoid blocking the UI. It also hampers composability of requests, because multiple chained requests will incur a penalty from jumping back and forth from the main queue to a background queue.

This breaking change removes the `dispatc_async` and leaves that decision to callers, as it's not possible to anticipate the correct decision to use in every situation.

### DIFF
--- a/Source/Models/APIClient.swift
+++ b/Source/Models/APIClient.swift
@@ -5,23 +5,25 @@ import Result
 public struct APIClient {
   private let requestPerformer: RequestPerformer
   private let deserializer: Deserializer
+  private let scheduler: Scheduler
 
-  public init(requestPerformer: RequestPerformer = NetworkRequestPerformer(), deserializer: Deserializer = JSONDeserializer()) {
+  public init(requestPerformer: RequestPerformer = NetworkRequestPerformer(), deserializer: Deserializer = JSONDeserializer(), scheduler: Scheduler = mainQueueScheduler) {
     self.requestPerformer = requestPerformer
     self.deserializer = deserializer
+    self.scheduler = scheduler
   }
 }
 
 extension APIClient: Client {
   public func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, SwishError> -> Void) -> NSURLSessionDataTask {
-    return requestPerformer.performRequest(request.build()) { result in
+    return requestPerformer.performRequest(request.build()) { [schedule = scheduler] result in
       let object = result
         >>- self.validateResponse
         >>- self.deserializer.deserialize
         >>- T.ResponseParser.parse
         >>- request.parse
 
-      onMain { completionHandler(object) }
+      schedule { completionHandler(object) }
     }
   }
 }
--- a/Source/Models/Scheduler.swift
+++ b/Source/Models/Scheduler.swift
@@ -1,0 +1,9 @@
+public typealias Scheduler = ((() -> Void) -> Void)
+
+public let immediateScheduler: Scheduler = { completion in
+  completion()
+}
+
+public let mainQueueScheduler: Scheduler = { completion in
+  dispatch_async(dispatch_get_main_queue(), completion)
+}
--- a/Tests/Tests/APIClientSpec.swift
+++ b/Tests/Tests/APIClientSpec.swift
@@ -191,6 +191,51 @@ class APIClientSpec: QuickSpec {
           expect(error).toEventually(equal(SwishError.ServerError(code: expectedCode, data: performer.data)))
         }
       }
+
+      describe("scheduling completion blocks") {
+        it("dispatches onto the main queue by default") {
+          let performer = FakeRequestPerformer(responseData: .JSON([:]))
+          let client = APIClient(requestPerformer: performer)
+
+          var isOnMain: Bool?
+          client.performRequest(FakeRequest()) { _ in
+            isOnMain = NSThread.isMainThread()
+          }
+
+          expect(isOnMain).toEventually(beTrue())
+        }
+
+        it("doesn't dispatch onto main queue when using the immediate scheduler") {
+          let performer = FakeRequestPerformer(responseData: .JSON([:]), background: true)
+          let client = APIClient(requestPerformer: performer, scheduler: immediateScheduler)
+
+          var isOnMain: Bool?
+          client.performRequest(FakeRequest()) { _ in
+            isOnMain = NSThread.isMainThread()
+          }
+
+          expect(isOnMain).toEventually(beFalse())
+        }
+
+        it("dispatches via a custom scheduler if set") {
+          var calledNoopScheduler = false
+          var completed = false
+
+          let noopScheduler: Scheduler = { _ in
+            calledNoopScheduler = true
+          }
+
+          let performer = FakeRequestPerformer(responseData: .JSON([:]), background: false)
+          let client = APIClient(requestPerformer: performer, scheduler: noopScheduler)
+
+          client.performRequest(FakeRequest()) { _ in
+            completed = true
+          }
+
+          expect(calledNoopScheduler) == true
+          expect(completed) == false
+        }
+      }
     }
   }
 }
--- a/Source/Extensions/GCD.swift
+++ b/Source/Extensions/GCD.swift
@@ -1,5 +1,0 @@
-import Foundation
-
-func onMain(f: dispatch_block_t)  {
-  dispatch_async(dispatch_get_main_queue(), f)
-}
--- a/Swish.xcodeproj/project.pbxproj
+++ b/Swish.xcodeproj/project.pbxproj
@@ -29,6 +29,10 @@
 		2C721E5C1BD5A7E800846414 /* NetworkRequestPerformer.swift in Sources */ = {isa = PBXBuildFile; fileRef = F806950D1B962C5300C61B4A /* NetworkRequestPerformer.swift */; };
 		2C721E5D1BD5A7E800846414 /* APIClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = F88ED06C1B96736B0069F56C /* APIClient.swift */; };
 		2C721E5E1BD5A83B00846414 /* Swish.h in Headers */ = {isa = PBXBuildFile; fileRef = F80695121B962C5300C61B4A /* Swish.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		4A05CC281CEFBA460076955E /* Scheduler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A05CC271CEFBA460076955E /* Scheduler.swift */; };
+		4A05CC291CEFBA4B0076955E /* Scheduler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A05CC271CEFBA460076955E /* Scheduler.swift */; };
+		4A05CC2A1CEFBA4C0076955E /* Scheduler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A05CC271CEFBA460076955E /* Scheduler.swift */; };
+		4A05CC2B1CEFBA4C0076955E /* Scheduler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A05CC271CEFBA460076955E /* Scheduler.swift */; };
 		E506EC7E1BB5BE380032E941 /* NimbleMatchers.swift in Sources */ = {isa = PBXBuildFile; fileRef = E506EC7D1BB5BE380032E941 /* NimbleMatchers.swift */; };
 		E51722EF1CEE5D1000B0C915 /* JSONDeserializer.swift in Sources */ = {isa = PBXBuildFile; fileRef = E58AD6361C91055200AD2CDE /* JSONDeserializer.swift */; };
 		E51722F01CEE5D1A00B0C915 /* Deserializer.swift in Sources */ = {isa = PBXBuildFile; fileRef = E58AD6321C91049300AD2CDE /* Deserializer.swift */; };
@@ -37,7 +41,6 @@
 		E52E5DA61CE7AF2500023C91 /* NSError.swift in Sources */ = {isa = PBXBuildFile; fileRef = F8D09FB81BFD360A00FB2433 /* NSError.swift */; };
 		E52E5DA71CE7AF2500023C91 /* NSURLRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = E5915B201BDABC4B005E5D63 /* NSURLRequest.swift */; };
 		E52E5DA81CE7AF2500023C91 /* Result.swift in Sources */ = {isa = PBXBuildFile; fileRef = F88ED0711B967C4A0069F56C /* Result.swift */; };
-		E52E5DA91CE7AF2500023C91 /* GCD.swift in Sources */ = {isa = PBXBuildFile; fileRef = F8D09FBB1BFD48F200FB2433 /* GCD.swift */; };
 		E52E5DAA1CE7AF2500023C91 /* SwishError.swift in Sources */ = {isa = PBXBuildFile; fileRef = F8F59BB51C2B34B60020B5BE /* SwishError.swift */; };
 		E52E5DAB1CE7AF2500023C91 /* RequestMethod.swift in Sources */ = {isa = PBXBuildFile; fileRef = F867938B1BD29E37007D9E98 /* RequestMethod.swift */; };
 		E52E5DAC1CE7AF2500023C91 /* NetworkRequestPerformer.swift in Sources */ = {isa = PBXBuildFile; fileRef = F806950D1B962C5300C61B4A /* NetworkRequestPerformer.swift */; };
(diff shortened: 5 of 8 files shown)


sidraval:
Yesssss yes yes, excellent find

gfontenot:
The original reasoning for doing this is to mimic the existing behavior of popular libraries like AlamoFire. As a user, it drives me _crazy_ to have to do this at the app level, especially since these kinds of requests _tend_ to require UI updates 80% of the time. I think completely removing the behavior might not be the best idea, _but_ I think injecting a specific queue to perform the completion block on might be interesting.

For example, something like:

``` diff
diff --git a/Source/Models/APIClient.swift b/Source/Models/APIClient.swift
index 83bba08..52c75c4 100644
--- a/Source/Models/APIClient.swift
+++ b/Source/Models/APIClient.swift
@@ -11,10 +11,10 @@ public struct APIClient {
 }

 extension APIClient: Client {
-  public func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, T.ResponseError> -> Void) -> NSURLSessionDataTask {
+  public func performRequest<T: Request>(request: T, completionQueue: dispatch_queue_t = dispatch_get_main_queue(), completionHandler: Result<T.ResponseObject, T.ResponseError> -> Void) -> NSURLSessionDataTask {
     return requestPerformer.performRequest(request.build()) { result in
       let object = (result >>- deserialize >>- request.parse).mapError(request.transformError)
-      onMain { completionHandler(object) }
+      dispatch_async(completionQueue) { completionHandler(object) }
     }
   }
 }
diff --git a/Source/Protocols/Client.swift b/Source/Protocols/Client.swift
index c5df687..0935756 100644
--- a/Source/Protocols/Client.swift
+++ b/Source/Protocols/Client.swift
@@ -3,5 +3,5 @@ import Argo
 import Result

 public protocol Client {
-  func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, T.ResponseError> -> ()) -> NSURLSessionDataTask
+  func performRequest<T: Request>(request: T, completionQueue: dispatch_queue_t, completionHandler: Result<T.ResponseObject, T.ResponseError> -> ()) -> NSURLSessionDataTask
 }
```

This still performs a call to `dispatch_async`, which might still mean it introduces latency? it also clutters up the function declaration but that seems fine since it's fairly nice and clean at the call site.

I don't know that this necessarily fixes the issue, though. I'd be interested to know if this change (in the PR, not my diff) alone is enough to solve the performance issues @sidraval has seen in his client project.

gfontenot:
Another thought: what's the _actual_ issue here? Is the basic problem that chaining multiple requests results in introduced latency? If so, maybe there's a better way we can specify request chains or request dependencies? `NSOperation`s have the concept of dependencies. Is this something we need/want?

sharplet:
I think this summarises the issues for me:
- Chaining multiple requests involves unnecessary hops to the main thread and back, introducing latency.
- Forcing execution to change to a different queue reduces composability with other abstractions that handle scheduling, such as ReactiveCocoa.

I like the idea of being able to inject a queue. What do you think about making the queue optional, so if one is not specified the completion block will be executed synchronously? The main queue could potentially also be the default.

gfontenot:
So you're suggesting something like:

``` diff
diff --git a/Source/Models/APIClient.swift b/Source/Models/APIClient.swift
index 930a54c..093e2d4 100644
--- a/Source/Models/APIClient.swift
+++ b/Source/Models/APIClient.swift
@@ -11,10 +11,14 @@ public struct APIClient {
 }

 extension APIClient: Client {
-  public func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, SwishError> -> Void) -> NSURLSessionDataTask {
+  public func performRequest<T: Request>(request: T, queue: dispatch_queue_t? = dispatch_get_main_queue(), completionHandler: Result<T.ResponseObject, SwishError> -> Void) -> NSURLSessionDataTask {
     return requestPerformer.performRequest(request.build()) { result in
       let object = result >>- deserialize >>- request.parse
-      onMain { completionHandler(object) }
+      if let queue = queue {
+        dispatch_async(queue) { completionHandler(object) }
+      } else {
+        completionHandler(object)
+      }
     }
   }
 }
diff --git a/Source/Protocols/Client.swift b/Source/Protocols/Client.swift
index 9104b9c..6af3cf2 100644
--- a/Source/Protocols/Client.swift
+++ b/Source/Protocols/Client.swift
@@ -3,5 +3,5 @@ import Argo
 import Result

 public protocol Client {
-  func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, SwishError> -> ()) -> NSURLSessionDataTask
+  func performRequest<T: Request>(request: T, queue: dispatch_queue_t?, completionHandler: Result<T.ResponseObject, SwishError> -> ()) -> NSURLSessionDataTask
 }
```

sharplet:
Yes, something like that. What do you think about a property on `APIClient` instead of a method parameter?

gfontenot:
I guess the two arguments would be:
- I want to make sure all requests through this client perform their callback on the same thread
- I want to make sure I'm able to control the client performs the callback on on a request by request basis

So that means that if we assign the queue at the instance level, we'd have to create two clients to be able to satisfy the second argument. Conversely if we pass the queue in, there's a little bit of duplication at the call site when performing the request.

I'm honestly not sure which makes more sense, but I feel like I'm still leaning towards injecting the queue at the point where we perform the request.

sharplet:
The only other issue I can think of with passing the queue in as a parameter to `performRequest` is that there's no way to specify a _different_ default than the main queue. Or put differently, request callbacks will always be performed on the main queue, unless I remember to change every single callsite to pass in `nil`.

For that reason I think I'd prefer a parameter on `APIClient`. What do you think? This may just be a difference of opinion, however.

---

FWIW, I tried to see if we could allow both. Theoretically, that would look something like this:

``` swift
public struct APIClient {
  var queue: dispatch_queue_t? = dispatch_get_main_queue()
}

extension APIClient: Client {
  public func performRequest<T: Request>(request: T, queue: dispatch_queue_t? = self.queue, completionHandler: Result<T.ResponseObject, SwishError> -> Void) -> NSURLSessionDataTask {
    // ...
  }
}
```

But alas, you can't access `self` in a parameter default.

gfontenot:
Yeah, I think it's just a difference of opinion. Maybe the rest of @thoughtbot/ios can break the tie?

tonyd256:
This is a tough call for me ... I can see what @sharplet wants and it makes sense; however, my 90% use case is that I want the call the main thread anyways so I like that swish does it for me. Would it be too much to have a second `performRequest` that did take a queue with a nil default?

What about a second completion block? `performRequest(request:, completionOnMain:)` and `performRequest(request:, completion:)`?

sharplet:
@tonyd256 I _think_ I'd prefer that to a single method call with a default parameter, because then they'll show up as two different autocomplete options in Xcode.

Although having said that, I'm now thinking that Xcode probably does the same thing for a single method with a default parameter!

An issue with having two versions that only differ by the label for the completion block is that when using trailing closure syntax it's now ambiguous which will be used (or maybe it just won't compile?).

sharplet:
Regarding our difference of opinion: I've pushed 321aa79 which implements the `Client`-based queue configuration for your review. It may be interesting to compare this to a method-based implementation, so that we can get a handle on the differences in API complexity and usage.

dazmuda:
Seriously good point about trailing closure syntax.

I am perfectly fine with the client based queue config since it doesn't require any config if you just want to default to the main queue.
And it accommodates your need to default to not the main queue.

So, my vote's for that.

dazmuda:
Though, it would be nice to also have a method that takes a queue that takes first priority over the default queue.

Because it would be a pain to have to set the default queue property often.

So I want both, yes. 😁

sharplet:
Also just pushed some tests for the queue handling stuff!

sharplet:
> Because it would be a pain to have to set the default queue property often.

In that case I'm thinking I'd probably instantiate a single `APIClient` and configure it once.

Tests/Tests/APIClientSpec.swift
```diff
@@ -191,6 +191,50 @@ class APIClientSpec: QuickSpec {
...
+          var client = APIClient(requestPerformer: performer)
+          client.queue = nil
+
+          var isOnMain: Bool?
+
+          client.performRequest(FakeRequest()) { _ in
+            isOnMain = NSThread.isMainThread()
+          }
+
+          expect(isOnMain).toEventually(beFalse())
```
gfontenot, line 217: I don't think this actually tests anything. `isOnMain` is `false` by default, so this can't fail. Would probably be better to default it to `true` and remove the optionality.

Source/Models/APIClient.swift
```diff
@@ -2,19 +2,24 @@ import Foundation
 import Argo
 import Result
 
-public struct APIClient {
+public struct APIClient: Client {
   private let requestPerformer: RequestPerformer
 
+  public var queue: dispatch_queue_t? = dispatch_get_main_queue()
```
gfontenot: I don't think this should be mutable. That would introduce the possibility of performing a request and changing the queue before it returns, resulting in unexpected behavior. Can we instead pass this in via `init`?

Source/Models/APIClient.swift
```diff
@@ -2,19 +2,24 @@ import Foundation
...
+
   public init(requestPerformer: RequestPerformer = NetworkRequestPerformer()) {
     self.requestPerformer = requestPerformer
   }
-}
 
-extension APIClient: Client {
   public func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, SwishError> -> Void) -> NSURLSessionDataTask {
-    return requestPerformer.performRequest(request.build()) { result in
+    return requestPerformer.performRequest(request.build()) { [queue] result in
```
gfontenot: I see, you're capturing here. I still think that leaving it mutable makes the API more complicated and I'd like to avoid that. If you need multiple queues, you can create multiple clients.

gfontenot:
I don't think we should add multiple `performRequest` methods. That's adding way more complexity around this than we need.

I'm _still_ not convinced that a property is the way to go here. I'm also a little worried that this is introducing complexity to the API based on speculation. Do we have any actual proof that this is an issue and that this is an actual fix? I thought I remembered that @sidraval said removing the queue stuff in the client app he was having performance issues in didn't result in any measurable performance improvement, but maybe I'm misremembering?

I'm very torn on this.

Tests/Tests/APIClientSpec.swift
sharplet, line 217: No, it's `nil` by default. The matcher will take care of the `nil` case, failing the test with a "use beNil() to match nils" message.

Source/Models/APIClient.swift
sharplet: Semantically, if you set the API client's queue, you've created a second client. However I agree that it's more cumbersome in practice to create a `var` and set a property than just pass the value into the initialiser. I've pushed a change for that.

sharplet:
> I'm also a little worried that this is introducing complexity to the API based on speculation.

FWIW, if we close this PR I'll still consider it a successful discussion. As far as API design goes, that's exactly what this PR is: _speculation!_ I think it's valuable to have something concrete to discuss.

So you raise a good point that so far my advocating for this hasn't been particularly concrete, and really my argument is more an appeal to principal. I personally believe that it's up to the _app_ to decide whether dispatching everything onto the main queue is the correct default. I'm not convinced that it _always_ is, and so I'm not convinced that Swish should force that policy on every app. In practice, perhaps it's super useful and the potential performance problems I'm citing are a non-issue. I'm totally happy if that's the case! I'd just feel more comfortable if this wasn't hard-coded.

sidraval:
I'm in agreement with @sharplet here.

@gfontenot I did say that, though to be clear it was a 30m punt where I just quickly hacked it together to see if was the only problem.

However, reading the code made way more sense, because previously my RAC requests would wrapped in an `onBackground { }` and it looked like the callback was also getting called on the background as well. Removing the main thread default meant my RAC requests looked like:

``` swift
(__request code__).observeOn(UIScheduler()).start {
```

which is more explicit, and clearer IMO.

tonyd256:
Pipe Dream:

What if there was a Background Monad and we had a function `runInBackground` similar to `runDB` where it encapsulates a transaction. You can pass one request to it and it will perform that request and return to the main thread. You could also compose many requests or other background tasks and sequence them so that they all run in order before they are returned to the main thread. This would help with chaining tasks that need background threading but still allow a final return to the main thread.

Probably not easy but maybe gets the brain juices flowing for something.

gfontenot:
Here's another option brought up by @jnutting that might actually be better than anything we've suggested so far:

``` diff
diff --git a/Source/Models/APIClient.swift b/Source/Models/APIClient.swift
index 930a54c..f924e1d 100644
--- a/Source/Models/APIClient.swift
+++ b/Source/Models/APIClient.swift
@@ -11,10 +11,11 @@ public struct APIClient {
 }

 extension APIClient: Client {
-  public func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, SwishError> -> Void) -> NSURLSessionDataTask {
+  public func performRequest<T: Request>(request: T, scheduler: (() -> Void) -> Void = onMain, completionHandler: Result<T.ResponseObject, SwishError> -> Void) -> NSURLSessionDataTask {
     return requestPerformer.performRequest(request.build()) { result in
       let object = result >>- deserialize >>- request.parse
-      onMain { completionHandler(object) }
+
+      scheduler { completionHandler(object) }
     }
   }
 }
diff --git a/Source/Protocols/Client.swift b/Source/Protocols/Client.swift
index 9104b9c..f35d966 100644
--- a/Source/Protocols/Client.swift
+++ b/Source/Protocols/Client.swift
@@ -3,5 +3,5 @@ import Argo
 import Result

 public protocol Client {
-  func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, SwishError> -> ()) -> NSURLSessionDataTask
+  func performRequest<T: Request>(request: T, scheduler: (() -> Void) -> Void, completionHandler: Result<T.ResponseObject, SwishError> -> ()) -> NSURLSessionDataTask
 }
```

So instead of passing in a queue, we just pass in a closure. And that closure can do whatever the hell it wants to do. It could just be `{ $0() }`, or it could do some real wacky shit. It also removes the concern about the `dispatch_async` vs `dispatch_sync` bit. It should _probably_ help with interaction with RAC schedulers as well, right? I'm not familiar enough with that lib to know for sure.

This seems _way_ more flexible than anything else we've talked about.

There's still the question about where to pass it, but I'll concede to the group since it seems like people prefer passing it at the time of initialization. That would also make this a non-breaking change as well.

sharplet:
I agree that the `scheduler` block seems like it may be the simplest change. What do you think about it having the type `((() -> Void) -> Void)?` so that callers can opt for _no_ scheduler (meaning invoke the block directly)?

sharplet:
FWIW, I've pushed a version that uses the `scheduler` block, and rebased on master. Let me know what you think!

Source/Models/APIClient.swift
```diff
@@ -13,15 +17,19 @@ public struct APIClient {
 }
 
 extension APIClient: Client {
-  public func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, SwishError> -> Void) -> NSURLSessionDataTask {
+  public func performRequest<T: Request>(request: T, scheduler schedule: ((() -> Void) -> Void)? = mainQueueScheduler, completionHandler: Result<T.ResponseObject, SwishError> -> Void) -> NSURLSessionDataTask {
```
gfontenot: Do you think the optional is worth it here? It seems like a bit of magic, where consumers could very easily do

``` swift
let directScheduler: (() -> Void) -> Void = { $0 }
```

and get the same behavior without adding complexity here or adding semantics to passing `.None`.

gfontenot:
Only remaining question is about the use of the optional for the scheduler. I'm also curious about if you wanted to move the scheduler to the `APIClient` instance. I feel like you felt stronger about that than I did and was ready to back down.

sharplet:
Ended up switching to the non-optional version of the scheduler, and introduced `immediateScheduler` and `mainQueueScheduler`, with the latter being the default. I also switched back to the property version, which I do prefer.

sharplet:
Also I just pushed a second commit that shares the `tvOS` and `watchOS` schemes, which oddly weren't already shared. That should fix CI.

Source/Models/Scheduler.swift
```diff
@@ -0,0 +1,9 @@
+public typealias Scheduler = ((() -> Void) -> Void)
```
gfontenot, line 1: nice

PR #70, opened 2016-04-25
Use SwishError as our final error type

gfontenot:
Originally, we were trying to let users model their error types by
implementing a `transformError` function on their `Request` objects that
knew how to go from the generic error encountered by Swish to a more
explicit error type controlled by the client application.

Unfortunately, this was a fairly naive approach that ended up causing
more trouble than it was worth.
- Given that there were so many different places for the request to
  encounter an error, it wasn't nearly as simple to model the errors as
  it was the success cases. This means that the original intent of the
  API was lost from the start.
- It became confusing to know when you should use `SwishError` and when
  you should use `NSError`. The boundary was fairly blurry and users
  encountered both regularly.
- Being completely generic about the error type at the `Client` level
  made it extremely difficult to wrap. We had been using this in client
  applications for things like global error messaging, where we'd wrap
  the `Client` and then post a notification if we encountered an error
  at any time in the request process. This became basically impossible
  once we needed to refer to a generic type like `T.ResponseError`,
  because we had absolutely no clues as to what that object was. Given
  the rest of the issues listed above, this seemed like needlessly
  complicated API with no real concrete benefit.

That being said, having `SwishError` around as an error type _is_ a
better solution than `NSError` in the long run. This error type gives
the end user _much_ more context about exactly what went wrong, and that
will give them the ability to do something better with it. Therefore, I
think it makes sense to move to an all-`SwishError` interface, rather
than reverting to `NSError`.

### DIFF
--- a/Source/Models/APIClient.swift
+++ b/Source/Models/APIClient.swift
@@ -11,9 +11,9 @@ public struct APIClient {
 }
 
 extension APIClient: Client {
-  public func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, T.ResponseError> -> Void) -> NSURLSessionDataTask {
+  public func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, SwishError> -> Void) -> NSURLSessionDataTask {
     return requestPerformer.performRequest(request.build()) { result in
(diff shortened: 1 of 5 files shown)


sidraval:
👍  This looks much easier to handle. Agreed on SwishError over NSError.

PR #75, opened 2016-05-06
Add documentation around basic usage, public protocols

sidraval:
(I've put `sr-remove-json-dependence` as the base branch just to give a reasonable diff)

`Documentation/Basics.md` has basic Swish usage
`Documentation/Protocols.md` has a heuristic explanation of the protocols, as well as a short discussion on the intention of each

cc @thoughtbot/ios @thoughtbot/copy-editing

### DIFF
--- a/Documentation/Advanced.md
+++ b/Documentation/Advanced.md
@@ -1,0 +1,88 @@
+# Advanced Usage #
+
+What if the endpoints you need to query don't return JSON? Perhaps you have an
+endpoint, `GET https://www.example.com/isAuthorized` that returns a `200 OK`
+with `"yes"` or `"no"` as a response body. We'll have to do a few things:
+
+- Define a custom `ResponseParser` to provide to our `Request` object, because
+the default one parses `JSON`.
+- Define a custom `Deserializer` to provide to our `APIClient`, because the 
+default one uses `NSJSONSerialization`.
+
+We'll start with the custom `Deserializer`. It will simply pass along the 
+`NSData` if it exists:
+
+```swift
+struct StringDeserializer: Deserializer {
+  func deserialize(data: NSData?) -> Result<AnyObject, SwishError> {
+    guard let data = data where data.length > 0 else {
+      return .Success(NSNull())
+    }
+
+    return .Success(data)
+  }
+}
+```
+
+Then, our custom `ResponseParser` will use `String(data:encoding:)` to
+transform the incoming `NSData` to a `String` representation:
+
+```swift
+struct StringParser: Parser {
+  typealias Representation = String
+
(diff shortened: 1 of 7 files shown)


Documentation/Advanced.md
```diff
@@ -0,0 +1,88 @@
...
+  typealias Representation = String
+
+  static func parse(j: AnyObject) -> Result<String, SwishError> {
+    guard let data = j as? NSData else {
+      let error = NSError.error("Bad data!")
+
+      return .Failure(.InvalidJSONResponse(error))
+    }
+
+    let error = NSError.error("Couldn't convert to string!")
```
nickvelloff, line 41: May or may not want to reference a Swish `NSError` `extension` in a client implementation example, could cause confusion.

Documentation/Advanced.md
```diff
@@ -0,0 +1,88 @@
+# Advanced Usage #
+
+What if the endpoints you need to query don't return JSON? Perhaps you have an
+endpoint, `GET https://www.example.com/isAuthorized` that returns a `200 OK`
+with `"yes"` or `"no"` as a response body. We'll have to do a few things:
+
+- Define a custom `ResponseParser` to provide to our `Request` object, because
+the default one parses `JSON`.
+- Define a custom `Deserializer` to provide to our `APIClient`, because the 
+default one uses `NSJSONSerialization`.
```
sharplet, line 10: What do you think about swapping the order of these two, to match the order those steps are performed when decoding a response?

Documentation/Advanced.md
```diff
@@ -0,0 +1,88 @@
...
+
+```swift
+struct StringParser: Parser {
+  typealias Representation = String
+
+  static func parse(j: AnyObject) -> Result<String, SwishError> {
+    guard let data = j as? NSData else {
+      let error = NSError.error("Bad data!")
+
+      return .Failure(.InvalidJSONResponse(error))
```
sharplet, line 38: This error has nothing to do with JSON—has `InvalidJSONResponse` been renamed?

Documentation/Advanced.md
```diff
@@ -0,0 +1,88 @@
...
+Then, our custom `ResponseParser` will use `String(data:encoding:)` to
+transform the incoming `NSData` to a `String` representation:
+
+```swift
+struct StringParser: Parser {
+  typealias Representation = String
+
+  static func parse(j: AnyObject) -> Result<String, SwishError> {
+    guard let data = j as? NSData else {
+      let error = NSError.error("Bad data!")
```
sharplet, line 36: How about a more exemplary error message? Maybe something like `"expected NSData, got \(j)"`?

PR #76, opened 2016-05-07
Update SwishError

sidraval:
- Remove `InvalidJSONResponse` since Swish supports other formats now
- Adds `InvalidDeserializeError` to catch errors in the `Deserializer`
- Adds `InvalidParseError` to catch errors in the `Parser`

### DIFF
--- a/Source/Models/SwishError.swift
+++ b/Source/Models/SwishError.swift
@@ -3,7 +3,8 @@ import Argo
 
 public enum SwishError {
   case ArgoError(Argo.DecodeError)
-  case InvalidJSONResponse(NSError)
+  case DeserializationError(NSError)
+  case ParseError(NSError)
   case ServerError(code: Int, data: NSData?)
   case URLSessionError(NSError)
 }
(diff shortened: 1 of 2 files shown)


Source/Models/SwishError.swift
```diff
@@ -3,7 +3,8 @@ import Argo
 
 public enum SwishError {
   case ArgoError(Argo.DecodeError)
-  case InvalidJSONResponse(NSError)
+  case InvalidDeserializeError(NSError)
+  case InvalidParseError(NSError)
```
gfontenot: The naming here makes it sound like the error is what's invalid. I wonder if there's a better name for these?

sidraval: Perhaps just `InvalidDeserialization` and `InvalidParse`?

tonyd256: I think someone did mention at one point to remove Error from the end of them all. I'd be in support of that.

gfontenot: How about `InvalidDeserialization` and `ParseError`?

PR #78, opened 2016-05-14
Add tvOS, watchOS targets

sidraval:
First time adding these targets, so a thorough review is appreciated. Everything appears to build, and tests pass on the new tv test target. I looked at Argo's setup as reference.


gfontenot:
Can we also update `bin/test` to exercise all 4 targets? You can see [an example of that in Argo](https://github.com/thoughtbot/Argo/blob/master/bin/test)

gfontenot:
This CI failure is expected because of #79 

👍

PR #88, opened 2016-06-20
Update dependencies

sharplet:
Updates all dependencies to the latest compatible version. This is also a workaround for the problem described in #87.


gfontenot:
👍

PR #90, opened 2016-06-28
Update Result dependency

nickvelloff:


gfontenot:
I'm not sure this is actually needed? At least for Carthage, 2.1.0 should be picked up with the `~> 2.0.0` constraint. Not sure about CocoaPods, but if not we should just change that constraint to `~> 2.0`.

sharplet:
Carthage defines `~>` as "compatible with", and thus `~> 2.0.0` will match any `2.x`. However I believe CocoaPods inherits a slightly different meaning from Bundler, where `~> 2.0.0` actually matches `2.0.x` and excludes 2.1+. For consistency I'd recommend they both change to `~> 2.0` where they should both match the same set of versions.

nickvelloff:
Correct this was needed for CocoaPods, and updated in Carthage for consistency. But, @sharplet approach is better than my original. Updated.

sharplet:
Apologies for the delay here. Is this still an issue for you? If not, I'd prefer to close this now that Swish 2.0 is available.

nickvelloff:
No this is old news...

PR #94, opened 2016-08-29
Fix NSError extension

zinthemoney:
Make NSError extension public so that advanced usage works.
Fix advanced usage doc.

### DIFF
--- a/Documentation/Advanced.md
+++ b/Documentation/Advanced.md
@@ -35,13 +35,13 @@ struct StringParser: Parser {
     guard let data = j as? NSData else {
       let error = NSError.error("Bad data!")
 
-      return .Failure(.InvalidJSONResponse(error))
+      return .Failure(.DeserializationError(error))
     }
 
     let error = NSError.error("Couldn't convert to string!")
     let string = String(data: data, encoding: NSUTF8StringEncoding)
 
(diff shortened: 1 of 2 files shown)


jakecraige:
Thanks for the PR. I'm not sure exposing the `.error` function makes sense because the error it exposes is in the Swish domain, and if you're extending it in your app, that's no longer the swish domain.

What do you think about inlining the error initialization in the advanced usage doc instead?

Documentation/Advanced.md
```diff
@@ -67,7 +67,7 @@ struct AuthorizedRequest: Request {
         return Result(false)
       default:
         let error = NSError.error("Unexpected response body!")
-        return Result(error: .InvalidJSONResponse(error))
+        return Result(error: .DeserializationError(error))
```
jakecraige, line 70: Good catch, since `.InvalidJSONReponse` doesn't exist anymore. I'd prefer to call this a `.ParseError` now since that's the step were in here. Thoughts?

zinthemoney: +1 on `ParseError`.

sharplet:
Swish 2.0 now implements `CustomNSError` protocol, so this is no longer necessary.

PR #92, opened 2016-08-10
Support Xcode 8

sharplet:
Updates dependencies and records Swift 2.3 migration in Xcode.

I couldn't find anything about Xcode 8 support on Circle, so this isn't being tested on CI yet.

PR #95, opened 2016-09-18
Swift 3.0

jakecraige:
Built off @sharplet's open PR for 2.3 support

Do the minimum set of changes to support Swift 3.

There are other improvements we can make like:
- using `Any` instead of `AnyObject` (JSONSerialization returns this now)
- Adopt a more Swify API by not ignoring so many first args.

Open question(s):
- Now that we have `URLRequest`, should we still support `NSURLRequest` and `NSMutableURLRequest`?


Eke:
Hello, any updates on this PR?

jakecraige:
@Eke We need to wait for #92 before merging this, but in the meantime you should be able to reference this branch `swift-3.0` in your dependency manager to use this version

Eke:
@jakecraige Thanks for info !

.gitmodules
```diff
@@ -1,12 +1,15 @@
 [submodule "Carthage/Checkouts/Result"]
 	path = Carthage/Checkouts/Result
-	url = https://github.com/antitypical/Result.git
+	url = https://0243d50631cefef6b22c779f29d6414b513d1654@github.com/antitypical/Result.git
```
sharplet: This seems a little weird to me, a sha as a username in the URL?

Source/Extensions/NSError.swift
```diff
@@ -5,33 +5,33 @@ private let swishDomain = "com.thoughtbot.swish"
...
-    info[NSLocalizedDescriptionKey] = messageForStatusCode(statusCode)
+    info[NSLocalizedDescriptionKey] = messageForStatusCode(statusCode) as AnyObject?
     info[NetworkErrorDataKey] = data
 
     return NSError(domain: swishDomain, code: statusCode, userInfo: info)
   }
 }
 
-private func userInfoFor(function: String, _ file: String, _ line: Int) -> [String: AnyObject] {
+private func userInfoFor(_ function: String, _ file: String, _ line: Int) -> [String: AnyObject] {
```
sharplet: I think we can go ahead and use `Any` here, should be able to remove most of the explicit type annotations.

Source/Extensions/NSURLRequest.swift
```diff
@@ -1,25 +1,18 @@
...
-    let jsonArray = json as? [AnyObject]
-    return jsonArray ?? [:]
-  }
-}
-
-public extension NSMutableURLRequest {
-  override var jsonPayload: AnyObject {
     get {
-      return super.jsonPayload
+      let json = httpBody.flatMap { try? JSONSerialization.jsonObject(with: $0, options: JSONSerialization.ReadingOptions(rawValue: 0)) }
```
sharplet: I think `options:` gets a default value of `0`, so you should be able to remove that parameter completely.

Source/Models/JSONDeserializer.swift
```diff
@@ -2,33 +2,32 @@ import Argo
...
+  func parseJSON(_ data: Data?) -> Result<AnyObject, NSError> {
+    guard let d = data , d.count > 0 else {
+      return .success(NSNull())
     }
 
     return Result(
-      try NSJSONSerialization
-        .JSONObjectWithData(d, options: NSJSONReadingOptions(rawValue: 0))
-    )
+      attempt: { try JSONSerialization.jsonObject(with: d, options: JSONSerialization.ReadingOptions(rawValue: 0)) }
```
sharplet: Same here about removing `options:` in favour of the default value.

Source/Models/NetworkRequestPerformer.swift
```diff
@@ -2,23 +2,23 @@ import Foundation
 import Result
 
 public struct NetworkRequestPerformer: RequestPerformer {
-  private let session: NSURLSession
+  fileprivate let session: URLSession
 
-  public init(session: NSURLSession = NSURLSession.sharedSession()) {
+  public init(session: URLSession = URLSession.shared) {
```
sharplet: Curious: would `.shared` work in this context?

Source/Models/NetworkRequestPerformer.swift
```diff
@@ -2,23 +2,23 @@ import Foundation
...
     self.session = session
   }
 }
 
 public extension NetworkRequestPerformer {
-  func performRequest(request: NSURLRequest, completionHandler: Result<HTTPResponse, SwishError> -> Void) -> NSURLSessionDataTask {
-    let task = session.dataTaskWithRequest(request) { data, response, error in
+  func performRequest(_ request: URLRequest, completionHandler: @escaping (Result<HTTPResponse, SwishError>) -> Void) -> URLSessionDataTask {
+    let task = session.dataTask(with: request, completionHandler: { data, response, error in
```
sharplet: Trailing closure here?

Source/Models/Scheduler.swift
```diff
@@ -1,9 +1,9 @@
-public typealias Scheduler = ((() -> Void) -> Void)
+public typealias Scheduler = ((@escaping () -> Void) -> Void)
```
sharplet, line 1: I feel like this annotation isn't in the right place. The closure that gets passed _into_ the scheduler shouldn't need to be escaping, right?

Tests/Fakes/FakeRequestPerformer.swift
```diff
@@ -2,56 +2,56 @@ import Foundation
 @testable import Swish
 import Result
 
-func serializeJSON(j: [String: AnyObject]) -> NSData? {
-  return try? NSJSONSerialization
-    .dataWithJSONObject(j, options: NSJSONWritingOptions(rawValue: 0))
+func serializeJSON(_ j: [String: AnyObject]) -> Data? {
+  return try? JSONSerialization
+    .data(withJSONObject: j, options: JSONSerialization.WritingOptions(rawValue: 0))
```
sharplet: Default options here

Tests/Tests/APIClientSpec.swift
```diff
@@ -11,11 +11,11 @@ class APIClientSpec: QuickSpec {
           it("passes the request's request object to the request performer") {
             let request = FakeRequest()
             let performer = FakeRequestPerformer(
-              responseData: .JSON([:])
+              responseData: .json([:])
             )
             let client = APIClient(requestPerformer: performer)
 
-            client.performRequest(request) { _ in }
+            _ = client.performRequest(request) { _ in }
```
sharplet: I think `performRequest()` should be annotated with `@discardableResult`.

sharplet (commented):
I say let's go ahead and switch to using `Any` instead of `AnyObject`. That should really tighten things up!

I think we can approach public API changes (like the first argument thing you mentioned) in a separate PR.

I think we should _not_ support `NSURLRequest` and `NSMutableURLRequest`.

Source/Models/Scheduler.swift
tonyd256: I'm not sure what @jakecraige's intention was here.

Source/Models/JSONDeserializer.swift
```diff
@@ -15,19 +15,19 @@ struct JSONDeserializer: Deserializer {
...
+  func parseJSON(_ data: Data?) -> Result<Any, NSError> {
     guard let d = data , d.count > 0 else {
       return .success(NSNull())
     }
 
     return Result(
-      attempt: { try JSONSerialization.jsonObject(with: d, options: JSONSerialization.ReadingOptions(rawValue: 0)) }
-    ).map { $0 as AnyObject }
+      attempt: { try JSONSerialization.jsonObject(with: d) }
+    ).map { $0 }
```
sharplet: Maybe this map isn't necessary anymore?

Source/Models/APIClient.swift
```diff
@@ -3,19 +3,21 @@ import Argo
...
     self.deserializer = deserializer
     self.scheduler = scheduler
   }
 }
 
 extension APIClient: Client {
-  public func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, SwishError> -> Void) -> NSURLSessionDataTask {
+  @discardableResult
+  public func performRequest<T: Request>(_ request: T, completionHandler: @escaping (Result<T.ResponseObject, SwishError>) -> Void) -> URLSessionDataTask {
+
```
sharplet: Extra newline here

Source/Models/JSONDeserializer.swift
```diff
@@ -2,33 +2,32 @@ import Argo
...
     return Result(JSON.init(j))
   }
 }
 
 private extension JSONDeserializer {
-  private func parseJSON(data: NSData?) -> Result<AnyObject, NSError> {
-    guard let d = data where d.length > 0 else {
-      return .Success(NSNull())
+  func parseJSON(_ data: Data?) -> Result<Any, NSError> {
+    guard let d = data , d.count > 0 else {
```
sharplet: Extra space before the comma

Source/Protocols/Client.swift
```diff
@@ -3,5 +3,6 @@ import Argo
 import Result
 
 public protocol Client {
-  func performRequest<T: Request>(request: T, completionHandler: Result<T.ResponseObject, SwishError> -> ()) -> NSURLSessionDataTask
+  @discardableResult
+  func performRequest<T: Request>(_ request: T, completionHandler: @escaping (Result<T.ResponseObject, SwishError>) -> ()) -> URLSessionDataTask
```
sharplet: Thoughts on renaming this to `perform()`?

Source/Protocols/RequestPerformer.swift
```diff
@@ -2,5 +2,6 @@ import Foundation
 import Result
 
 public protocol RequestPerformer {
-  func performRequest(request: NSURLRequest, completionHandler: Result<HTTPResponse, SwishError> -> Void)-> NSURLSessionDataTask
+  @discardableResult
+  func performRequest(_ request: URLRequest, completionHandler: @escaping (Result<HTTPResponse, SwishError>) -> Void)-> URLSessionDataTask
```
sharplet: Same here about renaming to `perform()`

Eke:
Hello, can you please update `podspec` dependencies for `Argo` and `Result` ?

circle.yml
```diff
@@ -4,9 +4,7 @@ machine:
 
 dependencies:
   override:
-    - git submodule init
-    - git submodule foreach --recursive 'git fetch --tags'
-    - git submodule update --recursive
+    - bin/setup
```
eliperkins, line 7: I wish it were possible to fav a commit. Because I would fav this series of submodule commits for sure.

Source/Models/JSONDeserializer.swift
tonyd256: oh yep forgot to remove that

Source/Protocols/Client.swift
tonyd256: ya, seems like that is a good Swift 3 api change

ren:
Hey folks, any update on this getting merged? Is there any work which could be contributed to this PR to get it closer to ready?

Source/Models/Scheduler.swift
sharplet: My bad, the closure may be scheduled asynchronously, so it makes sense for it to be `@escaping` here.

sharplet:
@jakecraige @tonyd256 I've pushed a few more commits removing some unnecessary casts and updating our error handling to use the new `LocalizedError` and `CustomNSError` protocols. I think this should be about ready to merge!

Source/Extensions/Result.swift
```diff
@@ -2,12 +2,12 @@ import Argo
 import Result
 
 public extension Result {
-  static func fromDecoded<T>(decoded: Decoded<T>) -> Result<T, SwishError> {
+  static func fromDecoded<T>(_ decoded: Decoded<T>) -> Result<T, SwishError> {
```
jakecraige: Since we're in Swift 3 now, what about changing this to `static func from<T>(decoded: ....`?

Source/Models/APIClient.swift
```diff
@@ -29,12 +30,12 @@ extension APIClient: Client {
 }
 
 private extension APIClient {
-  func validateResponse(httpResponse: HTTPResponse) -> Result<NSData?, SwishError> {
+  func validateResponse(_ httpResponse: HTTPResponse) -> Result<Data?, SwishError> {
```
jakecraige: Same here about Swift 3 API guidelines

Source/Models/SwishError.swift
```diff
@@ -2,47 +2,69 @@ import Foundation
...
+extension SwishError: CustomNSError {
+  public static let errorDomain = "com.thoughtbot.swish"
+
+  public var errorCode: Int {
+    switch self {
+    case .argoError:              return 1
+    case .deserializationError:   return 2
+    case .parseError:             return 3
+    case .serverError:            return 4
+    case .urlSessionError:        return 5
```
jakecraige: I don't think we have a guideline for Swift, but generally we prefer not to align things like these `returns` because they can cause a lot of diff noise if we ever change the name of the longest one or add a longer one and all the indenting has to be updated

Source/Models/SwishError.swift
```diff
@@ -2,47 +2,69 @@ import Foundation
...
+    }
+  }
+
+#if swift(>=3.1)
+#else
+  public var errorUserInfo: [String: Any] {
+    var userInfo: [String: Any] = [:]
+    if let errorDescription = errorDescription {
+      userInfo[NSLocalizedDescriptionKey] = errorDescription
+    }
```
jakecraige: I think you can remove this unwrapping and set it explicitly because [`[]` accepts an optional](https://developer.apple.com/reference/swift/dictionary/1540848-subscript)?

bin/setup
```diff
@@ -1,4 +1,3 @@
 #!/usr/bin/env bash
 
-brew install carthage 2> /dev/null
```
jakecraige, line 3: 👏

Source/Models/SwishError.swift
sharplet: Works for me.

Source/Extensions/Result.swift
sharplet: Good idea. I refactored it to be an initailiser in c5e4917.

Source/Models/APIClient.swift
sharplet: Works for me.

Source/Models/SwishError.swift
sharplet: Removing the aligment felt pretty messy and hard to scan for me. I ended up putting the returns statements on separate lines instead.

jakecraige:
👍  (can't approve through review since GH says this is "my" PR)

PR #98, opened 2016-12-28
Perform two requests concurrently and combine their results

sharplet:
Adds a new extension to the `Client` protocol that allows two requests to be performed concurrently, with their results combined into a tuple:

```
perform(_: RequestA, _: RequestB, completionHandler: (Result<(ResponseA, ResponseB), SwishError>) -> Void)
```

If either fails, the other request is cancelled and the failure result is passed to the completion handler. A new `Cancelable` protocol allows cancelable things to be composed together (e.g., dispatch blocks and URL session tasks).

I plan to add some tests for the basic behaviours, especially around how cancellation is handled. I honestly am not sure how to write a test for the thread-safety aspects of this code, but I'm pretty sure all reads and writes to the combined state are appropriately synchronised. I'll do another review to make sure there's no race between updating the combined state and waiting on the request group to complete.

### DIFF
--- a/Source/Models/RequestGroup.swift
+++ b/Source/Models/RequestGroup.swift
@@ -1,0 +1,42 @@
+import Result
+
+final class RequestGroup<State>: Cancelable {
+  let client: Client
+  private var state: State
+  private var requests = CompositeCancelable()
+  private let group = DispatchGroup()
+  private let queue = DispatchQueue(label: "com.thoughtbot.swish.RequestGroup<\(State.self)>")
+
+  init(client: Client, state: State) {
+    self.client = client
+    self.state = state
+  }
+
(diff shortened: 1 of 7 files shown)


sharplet:
I also had to update the macOS deployment target to 10.10 to make use of `DispatchWorkItem`.

Source/Models/RequestGroup.swift
```diff
@@ -0,0 +1,42 @@
+import Result
+
+final class RequestGroup<State>: Cancelable {
+  let client: Client
+  private var state: State
+  private var requests = CompositeCancelable()
+  private let group = DispatchGroup()
+  private let queue = DispatchQueue(label: "com.movetogive.MTGClient.RequestGroup<\(State.self)>")
```
sharplet: Oops, accidentally let this slip in

PR #111, opened 2018-01-23
Add support for building with SwiftPM

klaaspieter:


klaaspieter:
Never mind. This only partially works and is generating a bunch of build errors. Unfortunately I don't have the time to fully figure this out right now :(

PR #113, opened 2018-03-02
Update documentation.

sidraval:
- Updates `Documentation/Basics.md` to include the removal of Argo, and working examples via `Codable`.
- Adds `SwishExamples.playground` that contains the code in `Documentation/Basics.md`.
- Documents the playground in `README.md` with instructions on how to use it. 

If you checkout this branch, the URL in `SwishExamples.playground` needs to be set to `https://raw.githubusercontent.com/thoughtbot/Swish/sr-update-documentation/Documentation/example.json` for the example to work; the URL that is in there now will work once this branch is merged to master.


Documentation/Basics.md
```diff
@@ -1,66 +1,57 @@
-# Basic Usage #
-
-Let's say you have an endpoint, `GET https://www.example.com/comments/1`, that returns the following JSON:
+Let's say we have an endpoint, `https://raw.githubusercontent.com/thoughtbot/Swish/master/Documentation/example.json`
```
sharplet: Should we leave `GET` in here?

Documentation/Basics.md
```diff
@@ -1,66 +1,57 @@
...
-  "id": 1,
+  "id": 4,
   "commentText": "Pretty good. Pret-ty pre-ty pre-ty good.",
   "username": "LarryDavid"
 }
 ```
 
-We'll model this with the following struct, and implement Argo's `Decodable` protocol to tell it how to deal with JSON:
+We will model the data with the following struct,
+and implement the `Codable` protocol so we can
```
sharplet: implement -> conform to

Documentation/Basics.md
```diff
@@ -1,66 +1,57 @@
...
-  }
+    func build() -> URLRequest {
+        let url = URL(string: "https://raw.githubusercontent.com/thoughtbot/Swish/master/Documentation/example.json")!
+        return URLRequest(url: url)
+    }
 }
 ```
 
-We can then use the Swish's default `APIClient` to make the request:
+We can then use the Swish's default APIClient to make the request:
```
sharplet: Do you think it makes sense to leave the backticks in here?

Documentation/Basics.md
```diff
@@ -1,66 +1,57 @@
...
-  let id: Int
 
-  func build() -> URLRequest {
-    let url = URL(string: "https://www.example.com/comments/\(id)")!
-    return URLRequest(URL: url)
-  }
+    func build() -> URLRequest {
+        let url = URL(string: "https://raw.githubusercontent.com/thoughtbot/Swish/master/Documentation/example.json")!
+        return URLRequest(url: url)
+    }
```
sharplet: Do you think it's necessary to reindent these examples?

README.md
```diff
@@ -6,17 +6,18 @@
...
 
 Here is the current Swift compatibility breakdown:
 
-| Swift Version | Swish Version |
-| ------------- | ------------ |
-| 3.X           | 2.X          |
-| 2.X           | 1.X          |
+| Swift Version | Swish Version  |
+| ------------- | ------------   |
+|           4.X | >= 2.0.3       |
```
sharplet: Should we bump the major version for this change?

SwishExamples.playground/Contents.swift
```diff
@@ -0,0 +1,56 @@
+import PlaygroundSupport
+import Swish
+import Result
+
+PlaygroundPage.current.needsIndefiniteExecution = true
+
+// Let's say we have an endpoint, https://raw.githubusercontent.com/thoughtbot/Swish/master/Documentation/example.json
```
sharplet: Would it be possible to use playground markup instead of regular comments? Or does it automatically render these as Markdown?

sharplet (commented):
Looking good 👍 just a few thoughts

README.md
sidraval: Definitely!

Documentation/Basics.md
sidraval: Nope!

Documentation/Basics.md
sidraval: Yep!

SwishExamples.playground/Contents.swift
sidraval: Good idea 👍

Documentation/Basics.md
```diff
@@ -1,60 +1,51 @@
...
-}
 
-extension Comment: Decodable {
-  static func decode(_ json: JSON) -> Decoded<Comment> {
-    return curry(Comment.init)
-      <^> j <| "id"
-      <*> j <| "commentText"
-      <*> j <| "username"
+  enum CodingKeys: String, CodingKey {
+    case id = "id"
```
sharplet: FWIW, you can leave off the explicit `= "id"` if the string value is the same as the case name. (Same goes for `username` below.)

README.md
```diff
@@ -66,6 +67,16 @@ target.
 
 ## Usage
 
+### Basic Playground
+
+You can see an example of Swish in action via the included `SwishExamples.playground`.
+
+To use that, clone this repository and run `carthage bootstrap`. When that
```
sharplet: Carthage used to build for all platforms by default — if this is still the case, could we add `--platform iOS` to match the platform the playground targets?

SwishExamples.playground/Contents.swift
```diff
@@ -0,0 +1,66 @@
...
+
+/*:
+We can model the request by defining a struct that
+implements Swish's `Request` protocol
+*/
+
+struct CommentRequest: Request {
+    typealias ResponseObject = Comment
+
+
```
sharplet: Extra newline

sharplet (approved):
Couple more comments, but :shipit:

PR #119, opened 2018-04-25
Fix building Swish 2.0 for Xcode 9.3

### LINKED ISSUE #118: Errors on 2.0.3 with Xcode 9.3
One of my projects is using Swish 2.0.3. I updated Xcode to 9.3 and I started to see a couple errors. It would be great to get a quick fix for Swish 2.0 out if possible. Unfortunately, we're not able to migrate to Swish 3.0 at this time. Here are the errors I'm seeing:
```
Swish/Source/Models/APIClient.swift:10:108: Struct 'JSONDeserializer' is internal and cannot be referenced from a default argument value
```
(can't use internal type in default arg)

and
```
Swish/Source/Protocols/Request.swift:29:21: Missing argument for parameter #1 in call
```
(needs another`()` inside `.success()`)

I've forked and branched from what looked like a good point for 2.0: 4c9233c. I don't want to open a PR onto master here -- is there a good branch I could PR with these updates? My changes are here:
https://github.com/gstro/Swish/tree/2.0-fixes . Could there be a 2.x release branch? Would love to have this updated on CocoaPods soon too.

gfontenot:
Xcode 9.3 brings along Swift 3.3, which has a couple _slightly_ different
behaviors that need to be special cased. Our current release (3.0+) of Swish
handles these changes already, but some users are unable to upgrade right now,
and we shouldn't make their lives more difficult.

This adds the fixes for Swift 3.3 without breaking compatibility for Swift 3.2
users.

I'm opening this PR for visibility but I'm going to delete the 2.0.3 branch as soon as I have a 2.0.3 release tagged.

Fixes #118

### DIFF
--- a/Source/Models/JSONDeserializer.swift
+++ b/Source/Models/JSONDeserializer.swift
@@ -1,8 +1,8 @@
 import Argo
 import Result
 
-struct JSONDeserializer: Deserializer {
-  func deserialize(_ data: Data?) -> Result<Any, SwishError> {
+public struct JSONDeserializer: Deserializer {
+  public func deserialize(_ data: Data?) -> Result<Any, SwishError> {
     let json = parseJSON(data)
 
     return json.analysis(
@@ -10,6 +10,8 @@ struct JSONDeserializer: Deserializer {
       ifFailure: { .failure(.deserializationError($0.error)) }
     )
   }
+
+  public init() {}
 }
 
 extension JSON: Parser {
--- a/Source/Protocols/Request.swift
+++ b/Source/Protocols/Request.swift
@@ -26,6 +26,10 @@ public extension Request where ResponseObject: Collection, ResponseObject.Iterat
 
 public extension Request where ResponseObject == EmptyResponse {
   func parse(_ j: JSON) -> Result<ResponseObject, SwishError> {
-    return .success()
+    #if swift(>=3.3)
+      return .success(())
+    #else
+      return .success()
+    #endif
   }
 }


Source/Models/JSONDeserializer.swift
```diff
@@ -1,15 +1,17 @@
 import Argo
 import Result
 
-struct JSONDeserializer: Deserializer {
-  func deserialize(_ data: Data?) -> Result<Any, SwishError> {
+public struct JSONDeserializer: Deserializer {
```
sharplet, line 4: Huh. Was this never public before?

Source/Protocols/Request.swift
```diff
@@ -26,6 +26,10 @@ public extension Request where ResponseObject: Collection, ResponseObject.Iterat
 
 public extension Request where ResponseObject == EmptyResponse {
   func parse(_ j: JSON) -> Result<ResponseObject, SwishError> {
+    #if swift(>=3.3)
```
sharplet: How do you feel about outdenting these?

sharplet (approved):
Minor comments, but lgtm 👍

Source/Models/JSONDeserializer.swift
gfontenot: It never needed to be! I was actually a little surprised too.

Source/Protocols/Request.swift
gfontenot: do you mean like

```swift
func parse() {
  #if swift(>=3.3)
    return .success(())
  #else
    return .success()
  #endif
}
```
?

gfontenot: I'm fine with it, just want to make sure we're on the same page wrt style.

sharplet: Oh I suppose I meant the C style, outdenting to the left edge:

```swift
  func parse() {
#if swift(>=3.3)
    return .success(())
#else
    return .success()
#endif
  }
```

gfontenot: That looks... real bad to me tbh. How do you feel about the way I just indented the file?

gfontenot: also this is the most bandaid-y of bandaid-y code so we probably don't need to go too deep into this here bikeshed.

sharplet: I'm not too fussed, so feel free to use whichever style you prefer. I think I tend to go for that style to minimise diff noise when the conditionals are eventually removed.

gfontenot: yep, I hear you. This is a dead-end refactor though, these changes are never even making it to master. I'm going to go ahead and merge this as is currently written.

PR #123, opened 2019-03-26
Support Swift 5.0

sgoodwin:
This adds support for Swift 5.0 and fixes a bug with the iSO6081 date formatter.


Source/Protocols/Request.swift
```diff
@@ -12,6 +11,8 @@ public protocol Request {
 
 public extension Request where ResponseObject: Decodable {
   func parse(_ data: Data) throws -> ResponseObject {
+    let decoder = JSONDecoder()
+    decoder.dateDecodingStrategy = .formatted(DateFormatter.iso8601Full)
     return try JSONDecoder().decode(ResponseObject.self, from: data)
```
gfontenot: Should this be using the `decoder` created above?

Swish.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
```
gfontenot, line 5: I think we want to ignore this file

gfontenot (approved):
Do we also need to make fixes for Swift 5 SPM?

sgoodwin:
We don't currently support SPM, do we?

sgoodwin:
Hey, wanna point out for the record: I can't merge this.

Source/Extensions/Result+Value.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Swish
+//
+//  Created by Samuel Goodwin on 3/26/19.
+//  Copyright © 2019 thoughtbot. All rights reserved.
+//
+
+import Foundation
+
+extension Result {
+  public var value: Success? {
```
sharplet: I don’t think Swish should be extending the standard library’s `Result` here. Given that this is already a breaking change, I think it would be fine to remove this extension.

Source/Protocols/Request.swift
```diff
@@ -12,11 +11,24 @@ public protocol Request {
 
 public extension Request where ResponseObject: Decodable {
   func parse(_ data: Data) throws -> ResponseObject {
-    return try JSONDecoder().decode(ResponseObject.self, from: data)
+    let decoder = JSONDecoder()
+    decoder.dateDecodingStrategy = .formatted(DateFormatter.iso8601Full)
```
sharplet: I can see the value in improving our default handling of ISO 8601 date formats, as well as other `JSONDecoder` configuration options. However, I think this could use some more discussion, and would prefer not to tie this to a Swift 5 update. Would you mind separating this change out into a separate PR, or an issue so we can discuss alternative designs?

Have you tried overriding this implementation of `parse(_:)` in your project? Did you run into any issues?

.gitignore
```diff
@@ -1,6 +1,9 @@
 # OS X Finder
 .DS_Store
 
+# Xcode workspace checks
+IDEWorkspaceChecks.plist
```
sharplet: I think it’s fine to check this file in rather than ignoring it.

sharplet (changes requested):
Sorry I didn’t see this sooner — thanks for the PR! I’ve left a few suggestions, but I’d love to get this merged soon.

Source/Protocols/Request.swift
sgoodwin: Sure, I'll move it to a different PR. I have added this to implementations of `parse(_:)`, but for most project this means overriding every single implementation as the default settings don't properly handle ISO8601 like they're supposed to.

Tests/Tests/APIClientSpec.swift
```diff
@@ -33,7 +32,14 @@ class APIClientSpec: QuickSpec {
...
 
-                expect(result?.value).toEventually(equal("hello world"))
+                func value(_ result: Result<String, SwishError>?) -> String? {
+                  if case .success(let value)? = result {
+                    return value
+                  }
+                  return nil
+                }
+                
+                expect(value(result)).toEventually(equal("hello world"))
```
sharplet: I believe this might work as `try? result?.get()`, or even more preferably `try result?.get()` if `expect()` will accept a throwing autoclosure. If the autoclosure form doesn’t work, then an explicit closure definitely should, e.g.: `expect { try result?.get() }`.

Tests/Tests/APIClientSpec.swift
```diff
@@ -66,8 +74,15 @@ class APIClientSpec: QuickSpec {
...
+                
+                func value(_ result: Result<EmptyResponse, SwishError>?) -> EmptyResponse? {
+                  if case .success(let value)? = result {
+                    return value
+                  }
+                  return nil
+                }
 
-                expect(result?.value).toEventually(beVoid())
+                expect(value(result)).toEventually(beVoid())
```
sharplet: Same note here about the shorthand using `result?.get()`.

Tests/Tests/APIClientSpec.swift
```diff
@@ -83,7 +98,14 @@ class APIClientSpec: QuickSpec {
...
 
-                expect(result?.value).toEventually(beVoid())
+                func value(_ result: Result<EmptyResponse, SwishError>?) -> EmptyResponse? {
+                  if case .success(let value)? = result {
+                    return value
+                  }
+                  return nil
+                }
+                
+                expect(value(result)).toEventually(beVoid())
```
sharplet: Same note here about the shorthand using `result?.get()`

Tests/Tests/APIClientSpec.swift
```diff
@@ -121,8 +143,10 @@ class APIClientSpec: QuickSpec {
           let client = APIClient(requestPerformer: performer)
           var error: SwishError?
 
-          client.perform(request) {
-            error = $0.error
+          client.perform(request) { result in
+            if case let .failure(resultError) = result {
```
sharplet: I think this pattern could also be tidied up with `result.get()`:

```swift
var result: Result<Foo, SwishError>?

client.perform(request) {
  result = $0
}

expect { try result?.get() }.toEventually(throwError(SwishError.serverError(code: expectedCode, data: performer.data)))
```

Tests/Tests/NetworkRequestPerformerSpec.swift
```diff
@@ -38,8 +37,10 @@ class RequestPerformerSpec: QuickSpec {
 
             let performer = NetworkRequestPerformer(session: fakeSession)
             performer.perform(exampleRequest()) { result in
-              returnedCode = result.value!.code
-              returnedData = result.value!.data
+              if case let .success(value) = result {
+                returnedCode = value.code
+                returnedData = value.data
```
sharplet: What do you think about this?
```swift
let value = try? result.get()
returnedCode = value?.code
returnedData = value?.data
```

Source/Protocols/Request.swift
```diff
@@ -12,7 +11,8 @@ public protocol Request {
 
 public extension Request where ResponseObject: Decodable {
   func parse(_ data: Data) throws -> ResponseObject {
-    return try JSONDecoder().decode(ResponseObject.self, from: data)
+    let decoder = JSONDecoder()
+    return try decoder.decode(ResponseObject.self, from: data)
```
sharplet: Would you mind inlining the `decoder` variable so that this no longer shows up in the diff?

sharplet (changes requested):
Thanks a bunch for updating, just a few more tweaks and then I think this will be ready to merge!

Tests/Tests/APIClientSpec.swift
sgoodwin: This one does not appear to work. The function given to expect is expected to produce a `String?`, so the `Predicate` fails to line up type-wise.

sharplet: Ah sorry I was using `Foo` as an example because I didn’t know what the actual type would be. This compiles for me:

```swift
          var result: Result<String, SwishError>?

          client.perform(request) {
            result = $0
          }

          expect { try result?.get() }.toEventually(throwError(SwishError.serverError(code: expectedCode, data: performer.data)))
```

sgoodwin: Ohh I see, the detail is you can expect to throw an error, not just expect to get an error our.

sidraval (approved):
Thanks @sgoodwin !

PR #124, opened 2019-06-07
Wholesale conversion to a Swift package

sidraval:
Futzin' around after watching the WWDC videos on SPM

Currently, this builds + tests just fine.

## TODO
- [x] SwiftPM support
- [x] Carthage support
- [x] CocoaPods support


gfontenot:
Most of this is moves, right? `s/Source/Sources`?

gfontenot:
fwiw, I think it's really early to be dropping CocoaPods/Carthage support. That being said, I think we can/should absolutely push our SwiftPM support forward as being the preferred method going forward. Once Xcode 10.11 fully launches I think I'd be 👍 on making the move final (especially since you can always get to the older releases on GitHub/Trunk)

sidraval:
@gfontenot –

Yeah it's mostly just moves + a `swiftformat` pass. (I should commit the `.swiftformat`, probably!) I totally agree that it is too early to drop CocoaPods/Carthage support, I just ignored them while getting the package set up for now – I intend to add them back in before merging.

sidraval:
I can't for the life of me get the tests building/passing on CI. I think that's the last thing we need before we can merge this.

.gitignore
```diff
@@ -35,3 +35,98 @@ Carthage/Build
 .idea/
 *.iml
 
+# Created by https://www.gitignore.io/api/swift,xcode,swiftpm
```
gfontenot: there must be duplication above, right?

gfontenot:
> I can't for the life of me get the tests building/passing on CI

Something about deps, right? Have you ran `bin/setup && bin/test` in a clean repo? I wonder if you have some old artifacts from previous builds kicking around.

.gitignore
sidraval: Yep, good catch, I cleanup up the whole `.gitignore` [here](https://github.com/thoughtbot/Swish/pull/124/commits/e768499a1500e5e7d734906b2b0fded15bfea3c0)

Package.swift
```diff
@@ -0,0 +1,27 @@
...
+let package = Package(
+    name: "Swish",
+    products: [
+        .library(
+            name: "Swish",
+            targets: ["Swish"]
+        ),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/Quick/Quick", .upToNextMajor(from: "2.0.0")),
```
gfontenot: Note that there's currently an issue with SPM that will cause these dependencies to be downloaded (but not compiled/linked) for all consumers of Swish. IMO, this is a deficiency in SPM and should be fixed there, but it's worth being aware of.

LICENSE
```diff
@@ -1,22 +0,0 @@
-Copyright (c) 2014 thoughtbot, inc.
```
sharplet: We should restore the license.

Documentation/Basics.md
```diff
@@ -1,58 +0,0 @@
-Let's say we have an endpoint, `GET https://raw.githubusercontent.com/thoughtbot/Swish/master/Documentation/example.json`
```
sharplet: Can we restore these docs?

Documentation/README.md
```diff
@@ -1,6 +0,0 @@
-# Documentation #
```
sharplet: Can we restore these docs?

Documentation/example.json
```diff
@@ -1,5 +0,0 @@
-{
```
sharplet: Can we restore these docs?

.gitignore
```diff
@@ -1,37 +1,109 @@
...
+
+### Xcode Patch ###
+*.xcodeproj/*
+!*.xcodeproj/project.pbxproj
+!*.xcodeproj/xcshareddata/
+!*.xcworkspace/contents.xcworkspacedata
+/*.gcno
+**/xcshareddata/WorkspaceSettings.xcsettings
 
+# End of https://www.gitignore.io/api/swift,xcode,swiftpm
```
sharplet, line 110: I think we should also ignore `/.swiftpm/` (there are a couple of files from that directory checked in).

Package.swift
```diff
@@ -0,0 +1,27 @@
...
+        .package(url: "https://github.com/Quick/Nimble", .upToNextMajor(from: "8.0.0")),
```
sharplet: If you like, these could both use the shortand `from: "8.0.0"` syntax instead of `.upToNextMajor(from:)`.

Package.swift
```diff
@@ -0,0 +1,27 @@
...
+    ],
+    targets: [
+        .target(
+            name: "Swish",
+            dependencies: []
```
sharplet: This parameter can be left off if you like.

Source/Extensions/NSURLRequest.swift
```diff
@@ -1,12 +0,0 @@
-import Foundation
```
sharplet: If you're interested in avoiding all the renaming, you could consider maintaining the existing directory structure and using [the optional `path:` attribute](https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescription.md#methods-5) in the package manifest.

Sources/Swish/Models/Scheduler.swift
```diff
@@ -1,9 +1,11 @@
...
 public typealias Scheduler = ((@escaping () -> Void) -> Void)
 
 public let immediateScheduler: Scheduler = { completion in
-  completion()
+    completion()
 }
 
 public let mainQueueScheduler: Scheduler = { completion in
-  DispatchQueue.main.async(execute: completion)
+    DispatchQueue.main.async(execute: completion)
```
gfontenot: We're changing the indentation settings?

Sources/Swish/Models/SwishError.swift
```diff
@@ -0,0 +1,58 @@
...
+        case .decodingError:
+            return 1
+        case .serverError:
+            return 4
+        case .urlSessionError:
+            return 5
+        }
+    }
+
+    #if !swift(>=3.1)
```
gfontenot: Do we still support <3.1?

Swish.xcodeproj/Nimble_Info.plist
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gfontenot, line 1: What's up with this file (and the Quick ones)?

Tests/LinuxMain.swift
```diff
@@ -0,0 +1,7 @@
+import XCTest
```
gfontenot: Does this need to go in the `SwishTests` directory?

bin/test
```diff
@@ -2,30 +2,4 @@
...
-  | xcpretty --color
-
-xcrun xcodebuild \
-  -workspace Swish.xcworkspace \
-  -scheme Swish-watchOS \
-  -destination "generic/platform=watchOS" \
-  CODE_SIGNING_REQUIRED=NO \
-  CODE_SIGN_IDENTITY="" \
-  | xcpretty --color
+swift test
```
gfontenot: should we kill `bin/test` completely? It's not doing much for us at this point.

gfontenot (approved):

Source/Models/APIClient.swift
```diff
@@ -1,44 +0,0 @@
-import Foundation
-
-public struct APIClient {
-  fileprivate let requestPerformer: RequestPerformer
-  fileprivate let scheduler: Scheduler
-
-  public init(requestPerformer: RequestPerformer = NetworkRequestPerformer(), scheduler: @escaping Scheduler = mainQueueScheduler) {
```
sharplet: Oh hey I think these files are showing up as deleted + added because they've also been reindented from 2 to 4 spaces. What do you think about going back to 2 space indent?

This .swiftformat file seems to have reasonable results:

```
--exclude Carthage
--indent 2
--ifdef outdent
--self init-only
```

sharplet (commented):
Left a few comments for you, let me know what you think. Really excited for this! The main one that would make it a little easier to review this would be changing back to a 2 space indent.

Sources/Swish/Models/SwishError.swift
sidraval: Nope – I'll remove this as part of a separate PR

Tests/LinuxMain.swift
sidraval: I don't think this file even needs to exist, since we're only supporting Apple platforms at the moment

Sources/Swish/Models/Scheduler.swift
sidraval: Fixed thanks to @sharplet 's `.swiftformat` recommendation below

Issue #125, opened 2019-06-18
Swift.DecodingError on success 200 with empty body


martias:
When performing a POST request with an optional `typealias ResponseObject = SendResponse?` a `decodingError` is returned for a response with code 200 and empty body which should be the success response. 
The "SendResponse" JSON content is returned only when the server wants to inform an error.

sidraval:
Thanks for the report, @martias! I'll have some time to look into this tomorrow.

sidraval:
@martias This should be fixed on `master` – not quite sure when we'll get a new release out, but it shouldn't be too long!

martias:
Thanks for the fix @sidraval . We can use it form from the commit hash until a new version is released! 🙂

martias:
Are you planning on doing a new release soon?

PR #126, opened 2019-06-21
Fix a bug where optional ResponseObject types produced an error

sidraval:
https://github.com/thoughtbot/Swish/issues/125

When a `Request`'s `ResponseObject` was declared to be `Optional<T>`
where `T: Decodable`, Swish would erroneously produce a, `DecodingError`
if the HTTP response had empty body data.

This commit specializes the default `parse` implementation into two variants:

- One for when `typealias ResponseObject = T?`
- One for when `typealias ResponseObject = T`

### DIFF
--- a/Source/Protocols/Request.swift
+++ b/Source/Protocols/Request.swift
@@ -10,7 +10,7 @@ public protocol Request {
 }
 
 public extension Request where ResponseObject: Decodable {
-  func parse<Wrapped>(_ data: Data) throws -> ResponseObject where ResponseObject == Optional<Wrapped> {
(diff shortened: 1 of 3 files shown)


Source/Protocols/Request.swift
```diff
@@ -10,7 +10,11 @@ public protocol Request {
 }
 
 public extension Request where ResponseObject: Decodable {
-  func parse(_ data: Data) throws -> ResponseObject {
+  func parse<Wrapped>(_ data: Data) throws -> ResponseObject where ResponseObject == Optional<Wrapped> {
```
gfontenot: Should we remove `throws` here as well?

gfontenot (approved):

PR #127, opened 2019-06-23
add a MockClient for help in testing code that makes network requests

benweitzman:
This PR defines a MockClient class which implements the Client protocol. The MockClient can be registered with requests/response pairs in three ways:

1. A static response that will be given for any request of the given type
2. A static response that will be given for requests that match exactly the given request
3. A more flexible response that can depend on the request

### DIFF
--- a/Swish.xcodeproj/project.pbxproj
+++ b/Swish.xcodeproj/project.pbxproj
@@ -5,10 +5,10 @@
    objects = {
       "Nimble::Nimble" = {
          isa = "PBXNativeTarget";
-         buildConfigurationList = "OBJ_140";
+         buildConfigurationList = "OBJ_145";
          buildPhases = (
-            "OBJ_143",
-            "OBJ_196"
+            "OBJ_148",
+            "OBJ_201"
          );
          dependencies = (
          );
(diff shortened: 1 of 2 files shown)


gfontenot:
This is cool! I actually always meant to create a parallel lib for testing (called Swoosh) but never really got around to it. I think this would be a great addition to a lib like that! I will say that I'd be personally opposed to adding testing infrastructure to the main library API though. I would be worried about it polluting the namespace in calling code, and causing confusion for people trying to write network stacks.

benweitzman:
Great name! Makes sense about adding to the main library. Would you want to make a new target under this repo? Or break it out into its own thing entirely?

gfontenot:
> Would you want to make a new target under this repo? Or break it out into its own thing entirely?

Originally (this idea is years old now), I wanted to make a new repo. It might actually still exist (privately) under the thoughtbot org, actually. That was mostly because Carthage doesn't (afaik) work well with multiple libs being vended by the same repo. Cocoapods supports that better with subspecs, and Swift Package Manager works pretty well with it, I think. Given that we're working on moving to Swift PM (#124), maybe it would make sense to bundle this in as another product in the repo? Might be kind of nice to keep everything colocated in here.

Anyone else have opinions on this?

sharplet:
I think vending multiple products via SwiftPM is a great idea.

gfontenot:
So maybe it might make sense to put a pin in this until we have the package fully set up? Once that's done, we can rework this.

sidraval:
See #129, for whatever reason I can't push to @benweitzman 's fork anymore, though it seems to have let me do it once...?

PR #128, opened 2019-06-28
Adjust Podspec for 4.0.0

seanlabastille:
It seems that the Podspec was not updated for 4.0.0 which caught me by surprised as I was trying to update my dependencies.

Since 4.0.0 uses the Swift standard library `Result` the dependency was dropped from the Cartfile but not the Podspec.


sidraval:
Ah, thank you @flufff42 !

PR #129, opened 2019-07-26
Add a MockClient for help in testing code that makes network requests

sidraval:
This recreates #127 because I fucked that branch/PR up somehow 🤷‍♀


Source/Testing/Matchers/AnyRequestMatcher.swift
```diff
@@ -0,0 +1,11 @@
+internal struct AnyRequestMatcher<T: Request>: RequestMatcher {
+  func match<S>(_ request: S) -> Result<S.ResponseObject, SwishError>? where S: Request {
+    if S.self == T.self {
+      return self.response as? Result<S.ResponseObject, SwishError>
+    } else {
+      return .none
+    }
+  }
```
sidraval: On first glance it looks like we can omit the `if`/`else` entirely from the function body here, instead writing:

```swift
func match<S>(_ request: S) -> Result<S.ResponseObject, SwishError>? where S: Request {
  return self.response as? Result<S.ResponseObject, SwishError>
}
```

Letting the optional cast take care of the check for us. @benweitzman does that look right to you?

In any case, I think it'd be good to write some tests for MockClient and its matchers.

